Issue #21, opened 2015-10-18
Convert uploaded resource image into either JPEG or PNG format


juyrjola:

akx:
What input file formats do we accept? Or, say, anything that ImageMagick could handle?

juyrjola:
Basically just the ones that Pillow handles. Photos should be JPEG, ground plans and maps should be PNG, but I'm not sure if it makes sense to enforce that.

akx:
Okiedoke. Should we store the original file too, though, for possible later (re)conversion?

I assume the use case here is that the images the API outputs are "web usable"?

juyrjola:
You can get the original image with the default image URL (such as here: http://api.hel.fi/respa/v1/resource/aurhijh62voa/).

You can request smaller images by appending ?dim=<width>x<height> to the image URL.

akx:
The issue description seems to imply that images should be converted to JPEG/PNG upon ingestion/upload though (meaning that the original data uploaded would be lost)?

Or did I misunderstand things there?

juyrjola:
Right. If the original image is not JPEG nor PNG, it should be converted. If it already is in either supported format, nothing should be done.

The `cropping` field contains user-specified information on how to crop the image into a square shape for thumbnails, but we're still undecided if we should support other aspect ratios than 1:1.

PR #33, opened 2015-10-22
Transcode non-webby images to PNG

akx:
Refs #21

### DIFF
--- a/resources/models.py
+++ b/resources/models.py
@@ -1,21 +1,21 @@
-import struct
 import base64
-import time
 import datetime
-import pytz
+import struct
+import time
+
 import arrow
-from django.contrib.gis.db import models
+import django.contrib.postgres.fields as pgfields
+import django.db.models as dbm
+import pytz
 from django.conf import settings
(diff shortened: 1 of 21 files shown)


resources/models.py
```diff
@@ -536,6 +542,40 @@ def save(self, *args, **kwargs):
...
+        try:
+            img = Image.open(self.image)
+            img.load()
+        except Exception as exc:
+            raise InvalidImage("Image %s not valid (%s)" % (self.image, exc)) from exc
+
+        if img.format not in ("JPEG", "PNG"):  # Needs transcoding.
+            # TODO: Maybe some heuristics about when it'd be a better idea to transcode to a lossy JPEG instead?
+            image_bio = BytesIO()
+            img.save(image_bio, format="PNG")
```
juyrjola: It might be better to decide the format based on `Image.type`. Types `main` and `other` would get a JPEG format, and `map` and `ground_plan` would get converted to PNG.

akx: Sure thing!

Issue #32, opened 2015-10-21
Implement typeahead filtering for resources and units


juyrjola:
Using Elasticsearch?

akx:
How about an API like

`/v1/unit/typeahead?q=query` and 

```json
{
  "results": [
    {"id": ..., "title": ...},
    ...
  ]
}
```

as the response?

Should it maybe be possible to (optionally) return the full RESTy entity for each result?

PR #37, opened 2015-10-26
Typeahead API

akx:
Refs #32 

image

### DIFF
--- a/resources/api/__init__.py
+++ b/resources/api/__init__.py
@@ -2,6 +2,7 @@ from .base import all_views
 from .resource import ResourceListViewSet, ResourceViewSet, PurposeViewSet
 from .reservation import ReservationViewSet
 from .unit import UnitViewSet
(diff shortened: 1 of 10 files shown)


resources/api/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+from rest_framework import routers
+
+
+class RespaAPIRouter(routers.DefaultRouter):
+    def __init__(self):
+        super(RespaAPIRouter, self).__init__()
+        self.registered_api_views = set()
+        self.register_views()
+        self.register("typeahead", TypeaheadViewSet, base_name="typeahead")
```
juyrjola: I'd prefer the endpoint is called `search`, so that we don't have to change the API when we implement the full-text search functionality.

akx: If you think that'll be the same view(set), then sure.

PR #38, opened 2015-10-28
Add slugs for Unit and Resource

tuomas777:
Added django-autoslug 1.9.3 and autoslug fields for Unit and Resource. The fields are populated from name field and are unique. Slugs for existing units and resources are generated in migrations.

Refs #24

### DIFF
--- a/resources/migrations/0016_auto_20151028_1653.py
+++ b/resources/migrations/0016_auto_20151028_1653.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import autoslug.fields
+import resources.models.utils
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('resources', '0015_auto_20151028_1648'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='resource',
+            name='slug',
+            field=autoslug.fields.AutoSlugField(editable=False, unique=True, populate_from=resources.models.utils.get_translated_name),
+        ),
+        migrations.AlterField(
+            model_name='unit',
+            name='slug',
+            field=autoslug.fields.AutoSlugField(editable=False, unique=True, populate_from=resources.models.utils.get_translated_name),
+        ),
+    ]
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,3 +17,4 @@ django-image-cropping
 easy_thumbnails
 -e git+https://github.com/City-of-Helsinki/django-helusers.git#egg=django-helusers
 django-allauth
+django-autoslug==1.9.3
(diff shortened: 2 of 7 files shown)


akx:
Hmm, which language are slugs populated from?

tuomas777:
Finnish

akx:
I don't see how that is enforced... Maybe a test case to prove that no matter the active language context, slugs get populated from Finnish names? :)

juyrjola:
Indeed. `name_fi` is the assuredly Finnish attribute. In fact, we should use the first configured language in `settings.LANGUAGES`.

tuomas777:
Changed slug fields to be populated using <code>get_translated()</code> from models/utils.py, which essentially returns a translation in the first language in <code>settings.LANGUAGES</code>

resources/migrations/0016_auto_20151028_1653.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import autoslug.fields
+import resources.models.utils
+
+
+class Migration(migrations.Migration):
```
akx: Is this second migration required? :o

tuomas777: Yeah, because the slug fields need to be unique, and it cannot be set in the first migration.

akx: Can you explain why it can't be set in the first migration?

The only thing that comes to mind is that all of the resources will have `""` as their slug before the RunPython migration, which would be a violation of that uniqueness constraint. You _could_ set `null=True, unique=True` on the slug field to work around that (as they'd then default to `NULL`, which is not a value that's considered unique).

In any case, the AlterField operations could be moved into the other migration, after the RunPython op, to have only one migration per feature.

tuomas777: Yes the field needs to allow empty values at first, and then not allow them after populating the field. The problem is that you should not have schema altering changes and RunPython stuff in the same migration when using PostgreSQL according to Django docs: 

"Thus, on PostgreSQL, for example, you should avoid combining schema changes and RunPython operations in the same migration or you may hit errors like OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events."

At first I tried combining the migrations into one, but that resulted in the error. Using null=True unique=True from the beginning might work, then the migration would need to change the uniqueness, but still that would be against the recommendations.

Actually I'm not even sure about keeping the RunPython operations in the same migration as the field creations, but that seems to work without problems. 

Django docs solves this situation with three migrations [https://docs.djangoproject.com/en/1.8/howto/writing-migrations/#migrations-that-add-unique-fields]

akx: Got it! :) I wasn't familiar with this PostgreSQL quirk. Thanks for the thorough explanation.

akx:
LGTM!

Issue #41, opened 2015-10-29
Implement a model for resource equipment


juyrjola:
It's a many-to-many relationship to Resource. Include following fields:
- name (set to translated in translation.py)

tuomas777:
aliases for equipment, for example "videotykki" for "dataprojektori"

tuomas777:
In the join table, add a HStoreField for custom attributes and a translated `description` field.

PR #53, opened 2015-11-05
Add resource equipment

tuomas777:
- Added Equipment model and M2M through model ResourceEquipment between it and resources, which has also additional description field and data hstore. 

- For admin UI, added Equipment and ResourceEquipment to their own pages, and also inlined ResourceEquipment in Resource's admin page.

- For API, Equipment has it's own resource at the api root. Resource equipments
are inlined in resources' "equipments" field alongside equipment name
and id.

- Fixed TraslatedModelSerializer language filtering to exclude empty string fields in addition to null ones.

- Added installation of postgres hstore extension to Travis configuration.

- Fixed Django Rest Framework 3.3.x compatibility by adding an own pagination class that implements the same functionality as previously.

Refs #41 #52

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,9 @@ Installation
 
 ```shell
 sudo -u postgres createuser -L -R -S respa
+sudo -u postgres psql -d template1 -c "create extension hstore;"
 sudo -u postgres createdb -Orespa respa
-sudo -u postgres psql respa
-  CREATE EXTENSION postgis;
+sudo -u postgres psql respa -c "CREATE EXTENSION postgis;"
 ```
 
 2. Run Django migrations and import data
--- a/resources/api/base.py
+++ b/resources/api/base.py
@@ -47,16 +47,12 @@ class TranslatedModelSerializer(serializers.ModelSerializer):
             for lang in LANGUAGES:
                 key = "%s_%s" % (field_name, lang)
                 val = getattr(obj, key, None)
-                if val is None:
+                if val in (None, ""):
                     continue
                 d[lang] = val
 
             # If no text provided, leave the field as null
-            for key, val in d.items():
-                if val is not None:
-                    break
-            else:
-                d = None
+            d = (d or None)
             ret[field_name] = d
 
         return ret
--- a/resources/models/resource.py
+++ b/resources/models/resource.py
@@ -13,11 +13,12 @@ from django.utils.translation import ugettext_lazy as _
 from image_cropping import ImageRatioField
 from PIL import Image
 from autoslug import AutoSlugField
+from django_hstore import hstore
 
 from resources.errors import InvalidImage
 
 from .base import AutoIdentifiedModel, ModifiableModel
-from .utils import get_translated, get_translated_name
+from .utils import get_translated, get_translated_name, DEFAULT_LANG
 
 
 class ResourceType(ModifiableModel, AutoIdentifiedModel):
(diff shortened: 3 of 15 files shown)


juyrjola:
Passing to @akx for code review.

README.md
```diff
@@ -12,9 +12,9 @@ Installation
 
 ```shell
 sudo -u postgres createuser -L -R -S respa
+sudo -u postgres psql -d template1 -c "create extension hstore;"
 sudo -u postgres createdb -Orespa respa
-sudo -u postgres psql respa
-  CREATE EXTENSION postgis;
+sudo -u postgres psql respa -c "CREATE EXTENSION postgis;"
```
akx: Come to think of it, wouldn't just
```
sudo -u postgres psql respa -c "CREATE EXTENSION postgis; CREATE EXTENSION hstore;"
```
do to avoid touching `template1`?

resources/api/base.py
```diff
@@ -47,15 +47,12 @@ def to_representation(self, obj):
             for lang in LANGUAGES:
                 key = "%s_%s" % (field_name, lang)
                 val = getattr(obj, key, None)
-                if val is None:
+                if val is None or val == "":
```
akx: More succinctly said as `if val in (None, "")`

resources/api/base.py
```diff
@@ -47,15 +47,12 @@ def to_representation(self, obj):
...
                     continue
                 d[lang] = val
 
             # If no text provided, leave the field as null
-            for key, val in d.items():
-                if val is not None:
-                    break
-            else:
+            if not d:
```
akx:
```python
d = d or None  # if value is falsy, let value be null
```
(same thing, though makes it more obvious that (False, (), [], {}, 0, 0.0, ""), etc. are converted)

resources/models/resource.py
```diff
@@ -86,6 +99,7 @@ class Resource(ModifiableModel, AutoIdentifiedModel):
     max_period = models.DurationField(verbose_name=_('Maximum reservation time'), null=True, blank=True)
 
     slug = AutoSlugField(populate_from=get_translated_name, unique=True)
+    equipments = models.ManyToManyField(Equipment, verbose_name=_('Equipments'), through='ResourceEquipment')
```
akx: English nitpick: `equipment` is both singular and plural, so even if this is a m2m field, its name should be `equipment`. :D

resources/models/resource.py
```diff
@@ -378,3 +392,33 @@ class Meta:
         verbose_name = _('resource image')
         verbose_name_plural = _('resource images')
         unique_together = (('resource', 'sort_order'),)
+
+
+class ResourceEquipment(ModifiableModel):
+    resource = models.ForeignKey(Resource, related_name='resource_equipments')
+    equipment = models.ForeignKey(Equipment, related_name='resource_equipments')
+    data = hstore.DictionaryField(null=True, blank=True)
```
akx: So -- and I guess this is the most controversial point of the CR -- uh, what's the use case for making this a hstore field? What sort of `data` will be storing here in the near future that couldn't be normalized as normal relational fields or stored in a regular [`JSONField` (that naturally can't be directly queried)](https://github.com/bradjasper/django-jsonfield).

Now to be clear, I'm not directly criticizing this technical decision, I'd just like to know the reasoning.

resources/models/resource.py
```diff
@@ -378,3 +392,33 @@ class Meta:
...
+
+    class Meta:
+        verbose_name = _('resource equipment')
+        verbose_name_plural = _('resource equipments')
+
+    def __str__(self):
+        return "%s / %s" % (self.equipment, self.resource)
+
+
+class EquipmentAlias(ModifiableModel, AutoIdentifiedModel):
```
akx: Would love to have a docstring for this class (the usage is not directly clear).

resources/models/resource.py
juyrjola: There is no concrete use case yet, but the idea is to enable flexible and fast filtering. We want to allow different departments in the City to set their own custom attributes, and they will then build their customized UIs on top of our API.

akx: Okiedoke!

tuomas777:
Updated with the suggested changes.

akx:
:+1: :)

PR #68, opened 2015-11-20
Fewer reservation limitations for admins

tuomas777:
Refactored reservation validation logic quite a bit.

Previously reservations were validated in
Resource.get_reservation_period(), which was run also in
Reservation.save(). This caused it to be executed twice when
reservations were created / updated via the API. One more problematic
consequence was that it wasn't able to do any user specific validations,
which are needed now as we should have different restrictions for normal
users and staff members. That old method was also a bit cryptic, and was
probably giving false results on non-trivial reservation cases.

The new implementation tries to address these issues. In it validations
are divided to model level and API level ones. The model level
validations are meant to be common for all resources and they are
located in Reservation.clean().

Model level validations check that:

 * begin time < end time
 * the reservation doesn't overlap other reservations
 * begin time and end time are correctly aligned to match time slots

API level validations check that:

 * normal users cannot make multi day reservations
 * normal users cannot make reservations outside opening hours
 * normal users cannot make reservations longer than max reservation
   period for the resource

In practise API level validations are handled in
Resource.validate_reservation_period(), which replaces the old
get_reservation_period(). One major difference is that while the old
method was also altering begin and end times to match time slots, this
new one does not do that, it just returns an error in case the times
don't match time slots.

Also note that Reservation.clean() is not called in Reservation.save(),
so not even model level validations are run by default when saving an
reservation. In case that is needed for some reason, calling clean()
could be added to save(), but then the validations would be executed
twice, and that cannot be easily fixed.

Changed also max reservation limit check to be not applied to staff 
members, and moved it to it's own method in Resource.

Refs #62

### DIFF
--- a/resources/models/reservation.py
+++ b/resources/models/reservation.py
@@ -3,10 +3,11 @@ import django.contrib.postgres.fields as pgfields
 from django.conf import settings
 from django.contrib.gis.db import models
 from django.utils.translation import ugettext_lazy as _
+from django.core.exceptions import ValidationError
 from psycopg2.extras import DateTimeTZRange
 
 from .base import ModifiableModel
-from .utils import get_dt, save_dt
+from .utils import get_dt, save_dt, is_valid_time_slot
 
 
 class ReservationQuerySet(models.QuerySet):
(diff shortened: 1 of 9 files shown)


resources/models/reservation.py
```diff
@@ -75,10 +76,27 @@ class Meta:
     def __str__(self):
         return "%s -> %s: %s" % (self.begin, self.end, self.resource)
 
+    def clean(self):
+        if self.end <= self.begin:
```
juyrjola: Should you call `super()` before doing the rest of validation (to make sure that `self.resource` is set etc.)?

juyrjola: Ah, never mind!

Issue #70, opened 2015-11-23
Include user name and email in reservations when admin


juyrjola:
When querying resource reservations as a resource admin, include a nested `user` object with name and email attributes.

PR #73, opened 2015-11-24
Add user info object to reservation data

tuomas777:
Changed user field in reservation data to an object containing name, email and uuid. It is only shown to staff members. 

The change also affects how reservation user is given in a request, it needs to be "user": {"uuid": <uuid>} instead of just "user": <uuid>. This is relevant after reservation user and request user have been separated.

Refs #70


tuomas777:
Changed 'uuid' field to 'id', minor tweaks

PR #76, opened 2015-11-26
Better separation of reservation user and request user

tuomas777:
Now staff members can create and edit reservations belonging to other
users by specifying the user in the reservation request. Normal users
can create / edit only their own reservations.

The implementation also adds a reservable check, normal users cannot
make a reservation to a non reservable resource.

Reservation filtering is also affected. From now on, every user will
see all reservations by default. If one wants to see only her own
reservations, she needs to use query parameter is_own=true. Is_own field
is also added to reservation endpoint data.

Refs #69


resources/api/reservation.py
```diff
@@ -106,25 +113,32 @@ def to_representation(self, instance):
...
             except ValueError:
                 raise exceptions.ParseError('invalid user UUID value')
-            return queryset.filter(user__uuid=user_uuid)
+            queryset = queryset.filter(user__uuid=user_uuid)
+
+        if not request.user.is_authenticated():
+            return queryset
+
+        is_own = request.query_params.get('is_own', '').lower()
+        if is_own == 'true':
```
juyrjola: `if is_own in ('true', 't', 'yes', 'y', '1'):`

resources/api/reservation.py
```diff
@@ -106,25 +113,32 @@ def to_representation(self, instance):
...
+            queryset = queryset.filter(user=request.user)
+        elif is_own == 'false':
```
juyrjola: `elif is_own is not None:`

PR #87, opened 2015-12-09
Reservation mails

tuomas777:
Users need to be notified when their reservations are altered by admins. Emails will now be sent to the users when admins create, modify, or delete reservations for them via the API.

Implemented a generic Django template based mail building system which should make it easy to define new email types using the same base message. The mails are constructed from Django templates living in /resources/templates/mail/ directory.

The Respa mailing is controlled by two settings in the Django settings file:
RESPA_MAILS_ENABLED is used to switch the mailing on / off
RESPA_MAILS_FROM_ADDRESS determines the from address all the mails will use.

The PR also includes somewhat unrelated enhancements to disallowed method testing, the methods are now tested for all user types instead of just staff.

Refs #72

### DIFF
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2015-12-07 11:54+0200\n"
+"POT-Creation-Date: 2015-12-09 10:19+0200\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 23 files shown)


akx:
During stuff I've done for Desucon, I've noticed that it's a _tremendously_ good idea to queue mail sends somehow if you're using Plain Old SMTP. (This probably won't be an issue for Mailgun or other services.)

The simplest way to do this in Django is to switch to the [django-mailer](https://pypi.python.org/pypi/django-mailer/) backend and add a cronjob that sends the mails out.

jussiarpalahti:
I'm not that well versed in Django's mail system nor django-mailer specifically, but shouldn't it be simple just to configure a regular mailer (procmail, postfix, etc) on server to accept mails from Django immediately and let it take care of SMTP relay duties?

akx:
@jussiarpalahti: I'm not that well versed in mailer configuration, and I really think I don't want to become versed. :D Hence that solution instead :)

jussiarpalahti:
I did this some years ago with Postfix I reckon, and Exim is an option, too:
https://www.digitalocean.com/community/tutorials/how-to-install-the-send-only-mail-server-exim-on-ubuntu-12-04

These solutions don't require that much configuration and are probably more robust, but we'd have to get Ansible script for this...

akx:
@jussiarpalahti Either way, it seems that something about the SMTP conversation process is pretty slow -- like seconds slow. But this is a little OT for this PR, I guess. Just thought I'd mention it.

jussiarpalahti:
Well, OT insofar if the problem lies in code or its environment ;)

Issue #86, opened 2015-12-07
Return resource reservations in Excel format


juyrjola:
Use case: employees want to print out a calendar of future reservations to put on a wall.

PR #90, opened 2015-12-14
Reservation excel files

tuomas777:
Reservation list and detail endpoints return Excel .xlsx files if
the request's Accept header is to

application/vnd.openxmlformats-officedocument.spreadsheetml.sheet

The .xlsx files contain the same reservations as the endpoints normally
return with fields unit (name), resource (name), begin and end times,
user (email) and comments. User and comments are only added if the
requester has the rights to see those.

Internally the implementation uses the same serializer as before which
makes things a bit more complicated, but it also makes it possible to
use the same field restriction logic for .xlsx files.

Added also resource filter to reservation endpoint.

Refs #86

### DIFF
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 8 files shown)


akx:
Does this also work with DRF's `?format=xlsx` convention? :)

tuomas777:
Yeah it does

akx:
Cool beans! :)

PR #113, opened 2016-03-31
Admin permissions and more or less related stuff

### LINKED ISSUE #112: Add permission system for officials

tuomas777:
- Added django-guardian based user permissions: in order to approve a reservation, one must have `can_approve_reservation` permission for that unit. Approvers can change reservation state between `REQUESTED`, `DENIED` and `CONFIRMED`. Normal staff members have the same rights as before (basically can do anything except for changing reservation state).

- Limited reservation visibility: normal user can see all her reservations, and other reservations
that are in state REQUESTED or CONFIRMED. Staff members can see all reservations.

- Added approver field to reservations.

- Changed reservation delete to set state to `CANCELLED` instead of actually deleting the object.

- Prevented paid and confirmed reservation modifying.

- Hide reservation extra fields if the reservation isn't a paid one or if it isn't own.

- Some other minor enhancements

Closes #112 #98 #99


resources/api/reservation.py
```diff
@@ -254,6 +276,21 @@ class ReservationViewSet(munigeo_api.GeoModelAPIView, viewsets.ModelViewSet):
...
+        queryset = super().get_queryset()
+        user = self.request.user
+
+        # staff members can see all reservations
+        if user.is_staff:
+            return queryset
+
+        # normal users can see only their own reservations and reservations that are confirmed or requested
+        filters = Q(state__in=(Reservation.CONFIRMED, Reservation.REQUESTED))
+        filters |= Q(user=user) if user.is_authenticated() else filters
```
juyrjola: This would be clearer:
```
if user.is_authenticated():
    filters |= Q(user=user)
```

resources/api/reservation.py
```diff
@@ -267,13 +304,21 @@ def perform_create(self, serializer):
             instance.send_created_by_admin_mail()
 
     def perform_update(self, serializer):
+        kwargs = {'modified_by': self.request.user}
         old_instance = self.get_object()
-        new_instance = serializer.save(modified_by=self.request.user)
+        new_state = serializer.validated_data.get('state', old_instance.state)
+        if new_state == Reservation.CONFIRMED and old_instance.state != Reservation.CONFIRMED:
```
juyrjola: This would be clearer:
```
if new_state != old_instance.state:
    ...
```

resources/api/reservation.py
```diff
@@ -267,13 +304,21 @@ def perform_create(self, serializer):
...
+        elif old_instance.state == Reservation.CONFIRMED and new_state != Reservation.CONFIRMED:
+            kwargs['approver'] = None
+        new_instance = serializer.save(**kwargs)
+
         if self.request.user != new_instance.user:
             new_instance.send_updated_by_admin_mail_if_changed(old_instance)
 
     def perform_destroy(self, instance):
-        instance.delete()
+        instance.state = Reservation.CANCELLED
```
juyrjola: Maybe do this (to prevent abuse through repeated deletions)?
```
if instance.state == Reservation.CANCELLED:
    return
```

resources/models/reservation.py
```diff
@@ -105,6 +107,16 @@ def get_end_tz(self, tz):
     def is_active(self):
         return self.end >= timezone.now()
 
+    def is_paid(self):
```
juyrjola: Not `is_paid` but `needs_manual_confirmation`

resources/api/reservation.py
tuomas777: Hmm, but then it doesn't work the same way? That is meant to deal only with changes to and from `CONFIRMED`

juyrjola: Right. You would then need to check for `CONFIRMED` in an inner `if`. `Reservation.set_state`might be a better location for this.

PR #114, opened 2016-03-31
Prevent manually confirmed reservation cancelling

### LINKED ISSUE #99: Do not allow Users to change or delete accepted paid Reservations

tuomas777:
Also moved the permission check to has_object_permission()

Closes #99

### DIFF
--- a/resources/api/reservation.py
+++ b/resources/api/reservation.py
@@ -112,12 +112,6 @@ class ReservationSerializer(TranslatedModelSerializer, munigeo_api.GeoModelSeria
         if not resource.can_make_reservations(request_user):
             raise PermissionDenied()
 
-        # reservations that need manual confirmation and are confirmed cannot be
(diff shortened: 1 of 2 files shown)


resources/api/reservation.py
```diff
@@ -249,9 +243,17 @@ def filter_queryset(self, request, queryset, view):
...
 class ReservationPermission(permissions.BasePermission):
     def has_object_permission(self, request, view, obj):
-        if request.method in permissions.SAFE_METHODS or obj.resource.is_admin(request.user):
+        if request.method in permissions.SAFE_METHODS:
             return True
-        return obj.user == request.user
+
+        # reservations that need manual confirmation and are confirmed cannot be
+        # modified or cancelled without reservation approve permission
+        if all([obj.need_manual_confirmation(),
```
juyrjola: Using `and` here would be clearer

PR #137, opened 2016-09-15
Favorite resource

### LINKED ISSUE #126: Implement favorite/unfavorite functionality to resource endpoint
Also implement filtering by "favorite" status

tuomas777:
Implemented favorite resource functionality for officials.

favoriting: *POST* `/resource/<id>/favorite/`
unfavoriting: *POST* `/resource/<id>/unfavorite/`

Also two fields are added: `favorite_resources` field to user endpoint, `is_favorite` to resource endpoint. It is possible to filter by the `is_favorite` field.

Closes #126


resources/api/resource.py
```diff
@@ -113,6 +116,12 @@ def get_user_permissions(self, obj):
     def get_required_reservation_extra_fields(self, obj):
         return REQUIRED_RESERVATION_EXTRA_FIELDS if obj.need_manual_confirmation else []
 
+    def get_is_favorite(self, obj):
+        request = self.context.get('request', None)
+        if not (request and request.user.is_authenticated()):
+            return False
+        return request.user.favorite_resources.filter(id=obj.id).exists()
```
juyrjola: This will result in one new SQL query per returned Resource. Maybe it could be done using Django annotations instead?

resources/api/resource.py
```diff
@@ -245,19 +254,42 @@ def filter(self, qs, value):
...
     need_manual_confirmation = django_filters.BooleanFilter(name="need_manual_confirmation", widget=DRFFilterBooleanWidget)
+    is_favorite = django_filters.MethodFilter(widget=django_filters.widgets.BooleanWidget())
 
     class Meta:
         model = Resource
-        fields = ['purpose', 'type', 'people', 'need_manual_confirmation']
+        fields = ['purpose', 'type', 'people', 'need_manual_confirmation', 'is_favorite']
+
+    def filter_is_favorite(self, queryset, value):
+        favorites = self.user.favorite_resources.values_list('id') if self.user.is_authenticated() else []
```
juyrjola: I think you could put the DB to work by using `queryset.filter(favorited_by=self.user)` instead of fetching the favorite ids separately.

resources/api/resource.py
```diff
@@ -339,5 +370,33 @@ def get_queryset(self):
         else:
             return self.queryset.filter(public=True)
 
+    @detail_route(methods=['post'])
+    def favorite(self, request, pk=None):
```
juyrjola: Can `pk` actually be `None` here?

resources/api/resource.py
```diff
@@ -339,5 +370,33 @@ def get_queryset(self):
...
+            raise PermissionDenied()
+
+        if user.favorite_resources.filter(id=resource.id).exists():
+            return response.Response(status=status.HTTP_304_NOT_MODIFIED)
+        else:
+            user.favorite_resources.add(resource)
+            return response.Response(status=status.HTTP_201_CREATED)
+
+    @detail_route(methods=['post'])
+    def unfavorite(self, request, pk=None):
```
juyrjola: The methods share so much code that it would be better to refactor them to call a common setter function.

resources/api/resource.py
tuomas777: Yeah good point, I think this can be easily solved by prefetching the favorite resources, adding that.

resources/api/resource.py
tuomas777: For some reason thought that won't work. It is indeed way better, changing.

resources/api/resource.py
tuomas777: Not sure, but that function signature is from DRF docs.

resources/api/resource.py
tuomas777: okey dokey

tuomas777:
Implemented the suggested changes.

PR #151, opened 2016-10-10
Reservation access code

### LINKED ISSUE #147: Implement access code support in Reservations
Add field `Resource.access_code_type` as a CharField with only one choice for now: `pin6`. Other choices might be added in the future. If left `None`, access codes are not used for reservations to this resource. 

Add field `Reservation.access_code` as a CharField. API validates it based on `access_code_type`. If the field is left `null`, a random value is generated.

tuomas777:
Implemented access code functionality for reservations: 

* Added `access_code_type`field to Resource with choices `"none"` and `"pin6"` and `access_code` field to Reservation. 

* If an access code isn't provided in the posted data when creating a reservation, one is generated automatically. The access code cannot be changed via API nor using admin UI after the reservation has been created.

* If an user doesn't have `can_view_reservation_access_code` permission or she isn't a staff member, access code field is excluded from reservation data in the API from other than her own reservations.

* Enabled [TokenAuthentication](http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication) to Reservation endpoint. It is meant to be used by external clients that need to view reservation access codes.

* Added a new email notification which is sent when a reservation with an access code is created. It contains the access code alongside the reservation data.

Closes #147


resources/models/resource.py
```diff
@@ -347,6 +377,12 @@ def can_make_reservations(self, user):
     def can_approve_reservations(self, user):
         return self.is_admin(user) and user.has_perm('can_approve_reservation', self.unit)
 
+    def access_code_enabled(self):
```
juyrjola: I would name this `is_access_code_enabled()` to be consistent with other methods (start with a verb).

tuomas777:
Implemented the suggested changes.

PR #158, opened 2016-10-27
Allow setting Units and Resources reservable only for X days in advance

### LINKED ISSUE #156: Add reservation restriction to the amount of possible reservation days for Users

tuomas777:
Added field reservable_in_advance field to Unit and Resource,
which can be used to set the amount of days the entity can be reserved
in advance. If a Resource does not have a value, the Unit's value is
used as a fallback.

In API there is also reservable_before field which determines the
exact datetime the entity can be reserved until.

Freezegun lib is added to keep tests deterministic.

Closes #156


resources/api/resource.py
```diff
@@ -112,6 +112,8 @@ class ResourceSerializer(TranslatedModelSerializer, munigeo_api.GeoModelSerializ
     required_reservation_extra_fields = serializers.SerializerMethodField()
     is_favorite = serializers.SerializerMethodField()
     generic_terms = serializers.SerializerMethodField()
+    reservable_beforehand_in_days = serializers.ReadOnlyField(source='get_reservable_beforehand_in_days')
+    reservable_before = serializers.ReadOnlyField(source='get_reservable_before')
```
juyrjola: Shouldn't this be `None` for resource admins?

PR #163, opened 2016-12-01
Add ReservationMetadataField and ReservationMetadataSet models

tuomas777:
ReservationMetadataSets determine which extra fields are supported and which are required for a resource. The migration creates a default metadata set named varaamo_need_manual_confirmation, which matches previous implementation's supported and required reservation extra fields, and adds those sets to resources that need a manual confirmation.


resources/api/reservation.py
```diff
@@ -87,15 +87,27 @@ def __init__(self, *args, **kwargs):
...
+            required = set()
+            if resource.need_manual_confirmation:
+
+                # manually confirmed reservations have some extra conditions
+                can_approve = resource.can_approve_reservations(self.context['request'].user)
+                if data.get('staff_event', False) and can_approve:
+                    required = {'reserver_name', 'event_description'}
+
+            supported = resource.get_supported_reservation_extra_field_names()
+            if not required:
```
juyrjola: I think it's clearer to move this right after `if resource:` and lose the `if`. Override `required` then as needed.

resources/api/resource.py
```diff
@@ -374,7 +370,8 @@ def filter_queryset(self, request, queryset, view):
 class ResourceListViewSet(munigeo_api.GeoModelAPIView, mixins.ListModelMixin,
                           viewsets.GenericViewSet):
     queryset = Resource.objects.select_related('generic_terms', 'unit', 'type')
-    queryset = queryset.prefetch_related('favorited_by', 'resource_equipment', 'purposes', 'images', 'purposes')
+    queryset = queryset.prefetch_related('favorited_by', 'resource_equipment', 'purposes', 'images', 'purposes',
+                                         'reservation_metadata_set')
```
juyrjola: `reservation_metadata_set` should be in the `select_related` block.

resources/models/reservation.py
```diff
@@ -119,9 +119,9 @@ def is_active(self):
     def need_manual_confirmation(self):
         return self.resource.need_manual_confirmation
 
-    def are_extra_fields_visible(self, user=None):
+    def are_extra_fields_visible(self, user):
         if not self.need_manual_confirmation():
```
tuomas777: Yeah, because we want to limit extra field visibility based on the user only on reservations that need manual confirmation.

resources/models/resource.py
```diff
@@ -391,13 +393,39 @@ def get_reservable_days_in_advance(self):
     def get_reservable_before(self):
         return create_reservable_before_datetime(self.get_reservable_days_in_advance())
 
+    def get_supported_reservation_extra_field_names(self):
+        if not self.reservation_metadata_set:
```
tuomas777: Yes it does, but we cannot use `metadata_set.values_list()` if `metadata_set` is `None`

PR #167, opened 2016-12-14
Feature reports

tuomas777:
Also adds `host_name` extra field to Reservation.


reports/api/day_events.py
```diff
@@ -97,18 +107,51 @@ def validate_unit(self, value):
...
         return unit
 
+    def validate_resource(self, value):
+        if not value:
+            return None
+
+        resource_ids = value.split(',')
+        resource_qs = Resource.objects.filter(id__in=resource_ids)
+
+        if resource_qs.count() != len(resource_ids):
```
juyrjola: I don't think this check is necessary. If invalid ids are given, the resources will just be missing from the response.

PR #175, opened 2017-03-24
Import also kirjastot.fi unit's opening hours

jussiarpalahti:
These are only available through libraries.fi V3 API


jussiarpalahti:
There's couple of special units that wont show up in V2 kirjastot API. They don't function as regular libraries do and only exist to contain their own opening hours.

PR #186, opened 2017-04-05
Add more filters to reservation endpoint

### LINKED ISSUE #185: Add more filters to reservation endpoint
- event_subject (string)
- host_name (string)
- reserver_name (string)
- resource_name (string)
- is_favorite_resource (boolean)

tuomas777:
Closes #185


tuomas777:
Updated, added user based reservation excluding, see the commit message for details.

resources/api/reservation.py
```diff
@@ -378,6 +379,44 @@ def filter_queryset(self, request, queryset, view):
         return queryset
 
 
+class ReservationFilterSet(django_filters.FilterSet):
```
akx: This should probably derive from `django_filters.rest_framework.FilterSet` (and then I think the `DRFFilterBooleanWidget` hack wouldn't be needed)?

tuomas777: Yup it probably should, good catch.

tuomas777:
Updated, added permission to view reservation extra fields (and those are now hidden by default)

PR #189, opened 2017-04-11
Add user_permissions field to Reservation endpoint

### LINKED ISSUE #188: Add user_permissions field to Reservation API endpoint
Start with permissions `can_delete` and `can_modify`.

tuomas777:
Contains can_modify and can_delete fields (which are the same for now)

Closes #188


resources/api/reservation.py
```diff
@@ -260,6 +263,14 @@ def to_representation(self, instance):
     def get_is_own(self, obj):
         return obj.user == self.context['request'].user
 
+    def get_user_permissions(self, obj):
```
akx: Wonder if it'd be worth to just `raise SkipField` if this'd return false, so as to save a chunk of bandwidth...

PR #192, opened 2017-04-21
Add available_between filter to Resource endpoint

### LINKED ISSUE #187: Add available_between filter to resource endpoint

tuomas777:
The filter expects two comma-separated datetimes and returns resources
that are open and free on the given range. Both datetimes must be on the
same day.

The implementation is currently pretty naive and doesn't scale for large amount
of resources.

Closes #187


resources/api/resource.py
```diff
@@ -317,9 +320,49 @@ def filter_is_favorite(self, queryset, name, value):
...
+            raise exceptions.ParseError("'%s' must be a timestamp in ISO 8601 format" % value)
+
+    def filter_available_between(self, queryset, name, value):
+        if len(value) != 2:
+            raise exceptions.ParseError('available_between takes exactly two comma-separated values.')
+
+        available_start = self._deserialize_datetime(value[0])
+        available_end = self._deserialize_datetime(value[1])
+
+        if available_start.day != available_end.day:
```
akx: You probably want to compare the `date`s, not the `day`s. (`2016-01-01T00:00:00,2044-10-01T00:00:00`)

resources/api/resource.py
```diff
@@ -317,9 +320,49 @@ def filter_is_favorite(self, queryset, name, value):
...
+            raise exceptions.ParseError('available_between timestamps must be on the same day.')
+
+        reservations = Reservation.objects.filter(end__gt=available_start).filter(begin__lt=available_end)
```
akx: Some comments would be nice here. I assume this is filtering reservations that _somehow_ fall into the period between `available_start` and `available_end`? Does this take all of the 4 cases into account? (I tried searching for the SO question that has an elegant solution to this, but couldn't immediately find it. I'll get back to this if I do.)

resources/api/resource.py
```diff
@@ -317,9 +320,49 @@ def filter_is_favorite(self, queryset, name, value):
...
+        queryset = queryset.exclude(reservations__in=reservations)
+
+        not_available_ids = []
```
akx: This smells like it should be a set.

resources/api/resource.py
```diff
@@ -317,9 +320,49 @@ def filter_is_favorite(self, queryset, name, value):
...
+        for resource in queryset:
+            opening_hours = resource.get_opening_hours(available_start, available_end)
```
akx: FWIW, I think it'd make for prettier code if this loop body was refactored into a function of its own that would just return `True`/`False` depending on availability, instead of having to use `continue`s.

Then `not_available_ids` could be computed simply by way of 

```python
not_available_ids = set(
    resource.id
    for resource in resources
    if not _is_available_for_blah_whatever(resource, available_start, available_end)
)
```
and the function would also be more easily testable

resources/api/resource.py
```diff
@@ -317,9 +320,49 @@ def filter_is_favorite(self, queryset, name, value):
...
+
+            if len(opening_hours) > 1:
+                # range spans over multiple days, assume resources aren't open all night and skip the resource
+                not_available_ids.append(resource.id)
+                continue
+
+            hours = next(iter(opening_hours.values()))[0]  # assume there is only one hours obj per day
```
akx: Yigh, that's one ugly expression. 😆

resources/api/resource.py
tuomas777: Heh yeah of course

resources/api/resource.py
tuomas777: Yes we need reservations that overlap with the range between `available_start` and `available_end`, not sure what is wrong with this solution.

resources/api/resource.py
tuomas777: It sure is, got any better alternatives? :)

resources/api/resource.py
akx: Ah, here. http://stackoverflow.com/a/143568/51685

akx: I'm not sure there was anything wrong with it, it was just lacking a comment, and the query itself is somewhat obscure to the untrained eye.

PR #206, opened 2017-05-25
Make respa_exchange streaming notifications work for realz

juyrjola:


respa_exchange/ews/notifications.py
```diff
@@ -27,42 +27,51 @@ class SubscribeRequest(EWSRequest):
     """
     version = 'Exchange2013'
 
-    def __init__(self, principal):
+    def __init__(self, principal, streaming=False):
```
akx: I don't see any place where this would be initialized with `streaming=False`. In the spirit of YAGNI, maybe it's worth it to just get rid of the `PullSubscriptionRequest` code here?

In any case I'd prefer `StreamingSubscribeRequest` and `PullSubscribeRequest` or similar, instead of an argument that makes the class behave rather entirely differently (as far as EWS is concerned)?

respa_exchange/ews/session.py
```diff
@@ -64,13 +56,33 @@ def soap(self, request, timeout=10):
...
+        resp = self.post(self.url, timeout=timeout, **self._prepare_soap(request))
+        resp.raise_for_status()
+        return self._process_soap_response(resp.content)
+
+    def soap_stream(self, request, timeout=10):
+        """
+        Send an EWSRequest by SOAP and stream the response.
+        """
+        resp = self.post(self.url, timeout=timeout, stream=True, **self._prepare_soap(request))
+        for data in resp.iter_content(chunk_size=None):
```
akx: Hmm. I assume we're assuming that the chunks returned are always complete XML documents?

respa_exchange/listener.py
```diff
@@ -69,21 +79,144 @@ def run(self):
                 if failures >= 10:
                     log.warn('Killing off %s, too many failures', self)
                     self.please_stop = True
+
         log.info('Event thread %s dying', self)
 
+    def stop(self):
+        self.please_stop = True
+        # TODO: Interrupt requests.post()?
+        # self.join()
```
akx: Dead code :(

respa_exchange/listener.py
```diff
@@ -283,7 +313,8 @@ def handle_events(self):
         changed_resources = set()
         while True:
             try:
-                event = self.events.get(block=False)
+                event = self.events.get(timeout=1)
+                self.events.task_done()
```
akx: `task_done` seems to be new in Python 3.4.4. It might be worth it to check for its existence before calling it...

respa_exchange/ews/session.py
juyrjola: They always were during my testing. I'll experiment with lxml's incremental parsing and see if it's doable without too much of a hassle.

PR #227, opened 2017-12-18
Changes for Hämeenlinna reservation system for berths

derega:
Haltu has a project going on with Hämeenlinna municipality and respa was used as the basis for it. These are the changes we made during that project so far. The project is still ongoing and more changes are coming.

This pull request is made to begin the review process how Hämeenlinna and Helsinki requirements can be fulfilled in respa together. This is now way ready for merge yet, it's just easier to review like this :)

--------------

The hmlvaraus system is pretty much contained in `hmlvaraus` Django app. It uses/inherits/overrides stuff from `respa`/`resources`/`users` Django apps also in this same repo. 

### New API endpoints

* The API is used by authenticated and non-authenticated users.
* API endpoints in original respa had constraints which were not compatible with hmlvaraus, most notably time allotments

### Background tasks with Celery

The project uses Celery now to run background tasks. 

### Login views/templates / django-allauth

This is specific to Hämeenlinna and should not be part of respa. The respa repo is organised so that we had to put these in the same repo.

### Data Import from Hämeenlinna systems

This is specific to Hämeenlinna and should not be part of respa. The respa repo is organised so that we had to put these in the same repo.

### Translations

This is specific to Hämeenlinna and should not be part of respa. The respa repo is organised so that we had to put these in the same repo.

------------------

# Ideas for the future

It should be possible to run different configurations of respa in production. There should not be anything specific to certain user (municipality) of the project in the repo which provides the functionality. Now this repo contains stuff specific to Helsinki and Hämeenlinna. Like user authentication (helusers) and scripts to import data. Feature flags are not a good idea but somehow it should be possible to change the following easily:

* Translations
* Add/remove API endpoints
* Add/remove Django apps
* How the service is run in production (which features are enabled)

One idea would be to split the repo so that Django apps producing the functionality are in separate repo and Django project having the configurations are in separate repo.


petterill:
As per discussions with the developers, merging features from hml-varaus to Respa master requires extensive rewrites. So closing this pull request.

PR #230, opened 2018-02-01
Ignore nightly python failures for travis

vikoivun:


jussiarpalahti:
@vikoivun, How would we track fails on nightly after this..?

PR #232, opened 2018-03-14
Add optional period to resource available between filter

tuomas777:
Added an optional third parameter to resource available between
filter, which if given is used to determine a minimum free slot length
that must exists in the main time range. Start and end times must be on
the same day.

Known issue: because of limits in the current opening time implementation,
it is not possible to have end time at midnight, so if there is a free
time slot just before midnight, it cannot be detected.

Also performance might cause problems if there is a very large number of resources.

### DIFF
--- a/resources/api/resource.py
+++ b/resources/api/resource.py
@@ -7,7 +7,7 @@ import pytz
 from arrow.parser import ParserError
 
 from django import forms
-from django.db.models import Q
+from django.db.models import Prefetch, Q
 from django.urls import reverse
 from django.contrib.gis.db.models.functions import Distance
 from django.contrib.gis.geos import Point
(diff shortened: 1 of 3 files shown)


SamuelSaarentola:
Reservation minimum time parameter is missing

resources/api/resource.py
```diff
@@ -343,28 +381,72 @@ def _is_resource_open(self, resource, start, end):
...
+
+            if self._is_resource_available(resource, available_start, available_end, reservations, period):
+                available_resources.add(resource.id)
+
+        return queryset.filter(id__in=available_resources)
+
+    @staticmethod
+    def _is_resource_available(resource, available_start, available_end, reservations, period):
+        opening_hours = resource.get_opening_hours(available_start, available_end, resource.prefetched_opening_hours)
+        hours = next(iter(opening_hours.values()))[0]  # assume there is only one hours obj per day
```
vikoivun: Is this true enough? I've heard rumblings of places with self-service hours and full-service hours… No. I've no idea if they are affect this level of abstraction. 😬

PR #273, opened 2018-05-25
Feature/travis build docker

jussiarpalahti:


vikoivun:
Another suggestion: have a script executed on interactive logins, that tells the loggee where the application resides. I'll try and check if this is possible.

jussiarpalahti:
@vikoivun Um, another after which suggestion..? Did you forget to submit your review?

These containers should login directly to Django project root when docker exec to shell.

vikoivun:
Yes, I did. 😁 Anyway. I tested this, and creating a .bashrc worked. I thought it would be something like:

> Hello there!
> This container is running respa
> The code was checked out from: (github link) at (iso date)
> the version at build time was abc from branch bcd
> The code is at /opt/respa, which should by your cwd now
>
> respa is run by the gunicorn process in this container
>
> Following build options were used:
> PASSION
> KIWI
> MANGO

Nothing really earth shattering, but would be nice.

PR #285, opened 2018-06-05
Resource create view

dvainio:


respa/.env
```diff
@@ -0,0 +1,3 @@
+DEBUG=1
+ALLOWED_HOSTS=*
+DATABASE_URL=postgis:///respa
```
suutari-ai: Env file shouldn't be committed.

resources/models/availability.py
```diff
@@ -142,14 +142,17 @@ def _check_closed(self):
     def clean(self, ignore_overlap=False):
         super(Period, self).clean()
 
+        if self.start is None or self.end is None:
+            raise ValidationError(_("You must set 'start' and 'end' fields."), code="empty_start_end")
```
JsseL: Hmm.. since the `start` and `end` fields don't have `blank=True`, shouldn't they already be required in forms? See the [docs of blank attribute](https://docs.djangoproject.com/en/1.11/ref/models/fields/#blank):
> If a field has blank=True, form validation will allow entry of an empty value. If a field has blank=False, the field will be required.

resources/models/availability.py
```diff
@@ -142,14 +142,17 @@ def _check_closed(self):
...
+
         if self.start > self.end:
             raise ValidationError("Period must start before its end", code="invalid_date_range")
 
-        self._validate_belonging()
         self._check_closed()
 
     def save(self, *args, **kwargs):
         ignore_overlap = kwargs.pop('ignore_overlap', False)
+        self._validate_belonging()
```
JsseL: Commit message just said there was problems with this, nothing about which kind of problems. Just wondering what effects this change might have since the validation should be done during the clean..

Could we have a look of those problems you had?

respa/urls.py
```diff
@@ -43,7 +44,7 @@
     url(r'^resource_image/(?P<pk>\d+)$', ResourceImageView.as_view(), name='resource-image-view'),
     url(r'^v1/', include(router.urls)),
     url(r'^v1/reservation/ical/(?P<ical_token>[-\w\d]+).ics$', ICalFeedView.as_view(), name='ical-feed'),
-    url(r'^$', RedirectView.as_view(url='v1/'))
+    url(r'^$', RedirectView.as_view(url='v1/')),
```
JsseL: This seems a bit unrelated change.. not a biggie, but better to separate unrelated changes/fixed into their own commits/PRs.

respa_admin/views/resources.py
```diff
@@ -0,0 +1,127 @@
...
+            self.request,
+            extra=self.extra_formsets,
+            instance=self.object
+        )
+
+        return self.render_to_response(
+            self.get_context_data(
+                form=form,
+                period_formset_with_days=period_formset_with_days,
+                resource_image_formset=resource_image_formset,
```
JsseL: Somehow I have the feeling that the period and resource image formset could be part of the same top level formset i.e. part of the resource form that could actually be a formset. But we can discuss this in other channels...

JsseL:
Seems good. Definitely quite complicated Django form, but I think you got it covered quite nicely. I think we can merge it ones you get the test fixed and have addressed the comments.

JsseL:
I'll merge this. Let's keep improving it from here...

JsseL:
Oops.. missed that you had placed forms into `respa_admin/views/forms.py`. I think it makes more sense to have them in `respa_admin/forms.py`. More conventional and they aren't views so no need to have them under views. I'll fix this one.

PR #288, opened 2018-06-05
Add script and instructions for Respa Admin static resources

suutari-ai:
Add a script for building Respa Admin static resources with Yarn.

Instruct Travis to build the resources with that script.

Add section to README about running that script when setting up a
development environment.

Also add a `.dockerignore` file for ignoring node_modules and *.pyc when
building the Docker image.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,12 @@ python:
   - '3.5'
   - '3.6'
   - '3.7-dev'
+node_js: "8.11"
 
 matrix:
   allow_failures:
   - python: 3.7-dev
-  
+
 cache: pip
 
 # Respa uses int4range (IntegerRangeField) so needs at least Postgres9.2
@@ -32,6 +33,7 @@ install: 'pip install codecov -r requirements.txt'
 
 before_script:
   - psql template1 -c 'create extension hstore;'
+  - ./build-resources
 
 script: pytest --cov . resources respa_exchange caterings
 
--- a/build-resources
+++ b/build-resources
@@ -1,0 +1,6 @@
+#!/bin/sh
+set -e
+rootdir=$(dirname "$0")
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -10,11 +10,12 @@ python:
   - '3.5'
   - '3.6'
   - '3.7-dev'
+node_js: "8.11"
```
jussiarpalahti: I think we should use lts/* node as is done in (some) of our other projects?

build-resources
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
jussiarpalahti: Perhaps rename the file to .sh

.travis.yml
suutari-ai: What do you mean? 8.11 is the latest LTS Node currently (check https://nodejs.org/en/).

build-resources
suutari-ai: It's an executable script. It's a common convention to not have file extensions on executable files. See, e.g. https://askubuntu.com/questions/503127/should-i-save-my-scripts-with-the-sh-extension#503129

(Yeah, I know that many people coming from Windows world don't do this correctly and that's why you so often see `.sh` suffix on exectables too.)

suutari-ai: or https://google.github.io/styleguide/shell.xml#File_Extensions

.travis.yml
suutari-ai: Ah... checked the docs of Travis and now I understand what you mean. Will update the PR...

suutari-ai:
@jussiarpalahti Should be OK now. I'll just merge this and if there's anything more to tune, let's do that on another PR. (This ain't going to master yet anyway.)

PR #290, opened 2018-06-05
RespaAdmin: Add debug-toolbar

JsseL:
Makes development a bit nicer.

### DIFF
--- a/respa/settings.py
+++ b/respa/settings.py
@@ -108,6 +108,15 @@ MIDDLEWARE = [
     'django.middleware.security.SecurityMiddleware',
 ]
 
+if DEBUG:
+    INSTALLED_APPS.append('debug_toolbar')
+    INTERNAL_IPS = [
+        '127.0.0.1',
+    ]
+    MIDDLEWARE = [
+        'debug_toolbar.middleware.DebugToolbarMiddleware',
+    ] + MIDDLEWARE
+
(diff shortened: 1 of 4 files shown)


respa/settings.py
```diff
@@ -108,6 +108,13 @@
     'django.middleware.security.SecurityMiddleware',
 ]
 
+if DEBUG:
+    INSTALLED_APPS.append('debug_toolbar')
+    INTERNAL_IPS = ['127.0.0.1', '172.20.0.1',]
```
suutari-ai: The 172.20.0.1 won't be the same on different machines.

JsseL: Ah.. sorry, forgot by accident. I guess we'll just have to leave it out and expect that people know to add their own IPs there.

suutari-ai:
Actually. This is still missing the required python package. Though that should not probably be installed when running the tests. Maybe change the tests to run with DEBUG=0?

JsseL:
@vikoivun Shouldn't tests be ran without DEBUG?

JsseL:
@vikoivun Found the problem now. We defaulted to `DEBUG=True` in settings. Changed that since it ain't good practice. Let's see if the tests pass now.

JsseL:
@suutari-ai @dvainio merge?

JsseL:
Scheisse.. accidentally closed. Well no problem there..

PR #296, opened 2018-06-11
respa_admin: Implement Tunnistamo login

suutari-ai:
Make the Respa Admin views require logged in user with staff status and
to redirect to Tunnistamo login page if such user is not yet logged in
the current session.

For local development, this needs either modifying helusers to use local
Tunnistamo or to configure a development client_id in the "helsinki"
socialaccount.SocialApp instance and that client_id to be registered to
real Tunnistamo with correct callback URL, i.e. something like
http://localhost:8000/accounts/helsinki/login/callback/

When above OAuth2 client_id stuff is sorted out and the login is
working, then set the `is_staff` flag on for the user that got created
when testing the login.  This allows the user to use the Respa Admin.

NOTE: It is also possible to login via regular Django Admin and use that
existing session for authenticating to Respa Admin.

### DIFF
--- a/respa_admin/urls.py
+++ b/respa_admin/urls.py
@@ -1,5 +1,6 @@
-from django.conf.urls import include, url
+from django.conf.urls import url as unauthorized_url
 
+from . import views
 from .views.resources import (
     admin_index,
     ResourceListView,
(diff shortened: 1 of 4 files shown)


JsseL:
PR doesn't include changes to README. From the PR description I understood that some tricks are needed to get started with development. README should instruct how to do all such tricks.

respa_admin/urls.py
```diff
@@ -1,11 +1,13 @@
-from django.conf.urls import include, url
+from django.conf.urls import url as unauthorized_url
 
 from . import views
+from .auth import admin_url as url
```
JsseL: Neat. 👍

PR #291, opened 2018-06-05
Docker support for multi container dev environment

JsseL:
With this setup developers should be able to get up and running
without a hassle.

PR #299, opened 2018-06-19
Docker support for multi container dev environment

JsseL:
With this setup developers should be able to get up and running
without a hassle.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,30 @@
+version: '3'
+services:
+  db:
+    container_name: respa-db
+    build: ./docker/postgres/
+    restart: always
+    environment:
+        POSTGRES_USER: respa
+        POSTGRES_PASSWORD: respa
+        POSTGRES_DB: respa
+    volumes:
+      - db_data:/var/lib/postgresql
+  api:
+    container_name: respa-api
+    build:
+      context: ./
+      dockerfile: ./docker/dev/Dockerfile
+    command: bash -c 'tail -f /dev/null'
+    volumes:
+      - .:/usr/src/app
+    environment:
+      - DEBUG=1
+      - DATABASE_URL=postgis://respa:respa@db/respa
+    ports:
+      - "8000:8000"
+    depends_on:
+      - db
+
+volumes:
+    db_data:
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -1,0 +1,16 @@
+FROM python:3.5
+
+WORKDIR /usr/src/app
+
+ENV APP_NAME respa
+
+RUN apt-get update && apt-get install -y libgdal1h postgresql-client-9.4
+
+COPY requirements.txt ./requirements.txt
+COPY dev-requirements.txt ./dev-requirements.txt
+
+RUN pip install --no-cache-dir -r dev-requirements.txt
+
+COPY . .
+
+RUN echo 'alias rs="python manage.py runserver 0:8000"' >> ~/.bashrc
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,11 +7,11 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 line_length = 120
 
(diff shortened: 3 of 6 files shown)


JsseL:
@vikoivun @jussiarpalahti What do you think about this. This should not reflect the way docker dev env is handled in linkedevents as @jussiarpalahti pointed out in here #291.

JsseL:
@tuomas777 or @suutari-ai have time to review this? Would be nice to get this merged.

tuomas777:
I'm no docker expert but LTGM

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+  db:
+    container_name: respa-db
+    build: ./docker/postgres/
+    restart: always
+    environment:
+        POSTGRES_USER: respa
+        POSTGRES_PASSWORD: respa
+        POSTGRES_DB: respa
+    ports:
+        - "5432:5432"
```
suutari-ai: This will share the database on publicly available port, which is a security issue. (Think cases when you're accessing a publicly available network with your development laptop.)

Anyway, is this really needed? It would still be possible to connect to the db server from the host via `docker exec` or by using the containers IP address (visible via `docker inspect`).

JsseL: Hmm.. I guess we could omit it. Will remove.

docker/dev/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+ENV APP_NAME respa
+
+RUN apt-get update && apt-get install -y libgdal1h postgresql-client-9.4
+
+COPY requirements.txt ./requirements.txt
+COPY dev-requirements.txt ./dev-requirements.txt
+
+RUN pip install --no-cache-dir -r dev-requirements.txt
+
```
igordavydsson: Cannot see where you are installing the regular `requirements.txt` 🤔

JsseL: See [this](https://github.com/City-of-Helsinki/respa/blob/master/dev-requirements.txt#L1).

petterill:
[Review PR #299 (Docker support for multi container dev environment)](https://trello.com/c/aQRJYpQP/24-review-pr-299-docker-support-for-multi-container-dev-environment)

docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+    container_name: respa-api
+    build:
+      context: ./
+      dockerfile: ./docker/dev/Dockerfile
+    command: bash -c 'tail -f /dev/null'
+    volumes:
+      - .:/usr/src/app
+    environment:
+      - DEBUG=1
+      - DATABASE_URL=postgis://respa:respa@db/respa
```
igordavydsson: so, how does it connect to `db` if db container does not have ports exposed? could not make it work this way personally in my project...

JsseL: Good question. Ports seems to be missing indeed, but for some unexplained reason it still works on my machine (perhaps I have something special in my env, which I can't remember any more). Due to it working, probably didn't notice this before.

Anyways, don't have time to debug this atm. Did you try adding the ports? Did it work then?

jussiarpalahti: Intra-compose communication happens using Docker's DNS and other network stuff.

Anyway, there's expose in the original Dockerfile: https://github.com/docker-library/postgres/blob/3f585c58df93e93b730c09a13e8904b96fa20c58/9.5/Dockerfile#L172

jussiarpalahti:
I don't use Compose. If y'all think this works I can merge this. Docker images here work fine.

docker-compose.yml
JsseL: Ah, well that explains. Thanks @jussiarpalahti for linking that 👍. @igordavydsson perhaps you had another problem then.

JsseL:
@jussiarpalahti This at least makes the "get up and running" really simple: `docker-compose up` and that's pretty much it. I used it and it worked on my machine ™️ 🤞 . So my vote for merging and saving time of future devs. ✋

PR #303, opened 2018-06-25
Create Resource View Tests

dvainio:
Implement tests for create and update functionality.


suutari-ai:
You should also configure Travis to run the new tests by adding `respa_admin` to this line: https://github.com/City-of-Helsinki/respa/blob/master/.travis.yml#L36

suutari-ai:
It seems that the tests don't pass currently:
```
respa_admin/tests/test_resources.py:23: TypeError
>       assert response.context['form'].errors == {}
E       TypeError: 'NoneType' object is not subscriptable
```
@dvainio Can you fix them?

suutari-ai:
@dvainio: OK, I found what the problem was: The tests didn't use authenticated Django Test Client so they failed on authentication error. Fixed them now.

PR #308, opened 2018-06-29
WIP: Resource Form Time Periods

dvainio:
Wire up time period and days formsets with template.
Add fix for saving images (id and name were missing from template code).

### DIFF
--- a/respa_admin/templates/common/_period_day.html
+++ b/respa_admin/templates/common/_period_day.html
@@ -1,0 +1,35 @@
+{% with forloop.counter0 as dayId %}
+    <div class="weekday-row input-row" id="period-day-{{ id }}-{{ dayId }}">
+
+        <div id="day-db-ids">
+            {{ day.period }}
+            {{ day.id }}
+        </div>
+
+        <span class="select-dropdown btn btn-primary">
+            {{ day.weekday }}
+            <span class="glyphicon glyphicon-chevron-down" aria-hidden="true"></span>
+        </span>
+
+{#        {{ day.instance.get_weekday_display }}#}
+
+        <div class="form-group">
+            <div class="input-row range-input">
+                {{ day.opens }}
+                <span class="glyphicon glyphicon-minus"></span>
+                {{ day.closes }}
+            </div>
+        </div>
+
+        {{ day.closed }}
+
+{#        <button class="btn copy-next" type="button">#}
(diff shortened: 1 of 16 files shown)


respa_admin/templates/common/_period_day.html
```diff
@@ -0,0 +1,35 @@
...
+        {{ day.closed }}
+
+{#        <button class="btn copy-next" type="button">#}
+{#            <span class="glyphicons glyphicons-more-windows icon-left" aria-hidden="true"></span>#}
+{#            Kopioi seuravaan#}
+{#        </button>#}
+{##}
+{#        <button class="btn add-time left-group" type="button" id="btn-add-new-day">#}
+{#            <span class="glyphicon glyphicon-plus-sign icon-left" aria-hidden="true"></span>Lisää toinen aika#}
+{#        </button>#}
```
JsseL: Avoid having commented out a.k.a "dead code" in the PRs.

respa_admin/templates/resources/create_resource.html
```diff
@@ -15,31 +15,15 @@
...
-                {{ period_form.as_p }}
-
-                {% if period_form.days %}
-                    {{ period_form.days.management_form }}
-                    {% for day in period_form.days %}
-                        {{ day.as_p }}
-                    {% endfor %}
-                {% endif %}
-            {% endfor %}
+{#            {% include "resources/form/_opening_hours.html" %}#}
```
JsseL: Dead code

respa_admin/templates/common/_period_day.html
suutari-ai: Actually dead code means code that is uncommented but never called. Anyway, both are bad dead code and commented code.

suutari-ai: Hmm.. seems that I was wrong too. :smile: According to Wikipedia [dead code](https://en.wikipedia.org/wiki/Dead_code) means code that is executed, but whose results are never used. Correct term for code that is never called is actually [Unreachable code](https://en.wikipedia.org/wiki/Unreachable_code).

suutari-ai:
As @JsseL pointed out, there's some commented code which should be removed. Also tests would have been nice, but overall very nice work. Good job, Daniel! :+1:

I will merge this now to be able to demonstrate it in `respa-admin` branch, but I expect the commented code to be removed in another PR.

PR #319, opened 2018-08-22
Add reservation created notification

### LINKED ISSUE #255: As a user, I want to be notified by email when I make an instant reservation
The customer should be notified by email about instant reservations as well.

Currently notifications are sent like this:
user makes a preliminary/pending reservations (staff & user)
approved preliminary reservation (user)
user makes an instant reservation (none)
instant & preliminary reservation is cancelled (user)

tuomas777:
* Added notification support for instant reservations.

* Changed the notification system to allow non existing reservation notification templates. Without this change created reservations would cause errors before notification template for instant reservations is created. The change also makes the notification system generally more flexible, as you can disable individual reservation notifications from use.

Closes #255


notifications/migrations/0004_add_reservation_created_notification.py
```diff
@@ -0,0 +1,20 @@
...
+        ('notifications', '0003_add_more_notification_types'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='notificationtemplate',
+            name='type',
+            field=models.CharField(choices=[('reservation_requested', 'Reservation requested'), ('reservation_requested_official', 'Reservation requested official'), ('reservation_cancelled', 'Reservation cancelled'), ('reservation_confirmed', 'Reservation confirmed'), ('reservation_created', 'Reservation created'), ('reservation_denied', 'Reservation denied'), ('reservation_created_with_access_code', 'Reservation created with access code'), ('catering_order_created', 'Catering order created'), ('catering_order_modified', 'Catering order modified'), ('catering_order_deleted', 'Catering order deleted'), ('reservation_comment_created', 'Reservation comment created'), ('catering_order_comment_created', 'Catering order comment created')], db_index=True, max_length=100, unique=True, verbose_name='Type'),
+        ),
+    ]
```
suutari-ai: Rather than adding a new migration, I would just edit the latest migration file which adds/modifies this field and add the new choice there, since it's really just a metadata change and there is no real need for a migration. What do you think?

PR #323, opened 2018-08-28
Dockerfile: Fix names of installed apt packages

suutari-ai:
Install `gdal-bin` instead of `libgdal1h` and install
`postgresql-client` instead of `postgresql-client-9.4`.

This fixes the Docker image to build again with a more recent python:3.5
image, because the packages are upgraded in the newer image and
therefore don't exists with those names with the version specified in
them.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /usr/src/app
 
 ENV APP_NAME respa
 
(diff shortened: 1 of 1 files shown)


vikoivun:
I'll take a look at how the package dependencies are structured.

PR #332, opened 2018-09-05
Make municipality available in reservation mail

suutari-ai:
Make the municipality id of the unit of the reservation available for
use in the notification template so that it is possible to differentiate
the content of the notification by municipality.


petterill:
Some context: This feature was thought to be a solution for sending themed notification emails based on which city the user is from (Helsinki/Espoo/Vantaa). But then we realized that the location of the resource is not enough - we need to know for example which UI version was used to make the booking. So this is PR is not needed anymore, unless we come up with another use for this.

JsseL:
@petterill Will close the PR then (will keep the branch in case this is deemed useful in future).

PR #335, opened 2018-09-06
Respa admin translations support

dvainio:


resources/models/unit.py
```diff
@@ -54,7 +54,7 @@ class Unit(ModifiableModel, AutoIdentifiedModel):
                                                                   null=True, blank=True)
 
     class Meta:
-        verbose_name = _("unit")
+        verbose_name = _("Unit")
```
suutari-ai: Why the case change? Model and field verbose names are usually all lower case.

respa_admin/tests/test_resource_forms.py
```diff
@@ -29,10 +29,7 @@ def test_period_formset_with_invalid_period_data(valid_resource_form_data):
     request = RequestFactory().post(NEW_RESOURCE_URL, data=data)
     period_formset_with_days = get_period_formset(request)
     assert period_formset_with_days.is_valid() is False
-    assert period_formset_with_days.errors == [{
-        '__all__': ["You must set 'start' and 'end' fields."],
-        'start': ['Tämä kenttä vaaditaan.'],
-    }]
+    assert period_formset_with_days.errors != []
```
suutari-ai: This is making the assert very weak. How about something like:
```python
error_list = [set(x.keys()) for x in period_formset_with_days.errors]
assert error_list == [{'__all__', 'start'}]
```

respa_admin/tests/test_resource_forms.py
```diff
@@ -42,12 +39,8 @@ def test_period_formset_with_invalid_days_data(valid_resource_form_data):
...
     period_formset_with_days = get_period_formset(request)
     assert period_formset_with_days.is_valid() is False
-    assert period_formset_with_days.errors == [
-        {'__all__': ['Tarkista aukioloajat.']},
-    ]
-    assert period_formset_with_days.forms[0].days.errors == [
-        {'weekday': ['Tämä kenttä vaaditaan.']},
-    ]
+    assert period_formset_with_days.errors != []
+    assert period_formset_with_days.forms[0].days.errors != []
```
suutari-ai: Same here. You should still check that the dict keys are exactly `__all__` and `weekday` in those cases.

respa_admin/tests/test_resource_forms.py
```diff
@@ -62,9 +55,7 @@ def test_create_resource_with_invalid_data_returns_errors(admin_client, empty_re
         'purposes': ['Valitse oikea vaihtoehto.  ei ole vaihtoehtojen joukossa.'],
         'type': ['Tämä kenttä vaaditaan.'],
     }
-    assert response.context['period_formset_with_days'].errors == [
-        {'__all__': ['Tarkista aukioloajat.']},
-    ]
+    assert response.context['period_formset_with_days'].errors != []
```
suutari-ai: Ditto.

PR #347, opened 2018-09-18
Resource form translations

dvainio:
Added functionality for hiding and showing the translateable fields in the Resource form. Implemented the logic for calculating how many fields each corresponding language has (displayed on the language buttons in the form).


respa_admin/static_src/js/resourceFormLanguage.js
```diff
@@ -0,0 +1,100 @@
...
+  return currentLanguage;
+}
+
+/*
+* Helper function for getting languages to hide
+* based on the the current language.
+* */
+function getLanguagesToHide(currentLanguage) {
+  let languages = getAllLanguages();
+  return languages.filter(language => {return currentLanguage !== language});
```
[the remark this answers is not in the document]
suutari-ai: Yep, there's a missing semicolon, but you could also use the expression syntax lambda. Also prefer const over let, i.e.:

```javascript
const languages = getAllLanguages();
return languages.filter((x) => (x !== currentLanguage));
```

suutari-ai:
Check the unit tests. They seem to assume that name is a required field, but for some reason it's not anymore. Shouldn't one of the name fields (name_fi, name_en, or name_sv) be required though?

respa_admin/templates/resources/form/_nav.html
```diff
@@ -12,18 +13,12 @@ <h2>
...
-                    <span class="badge">5</span>
-                </button>
-                <button name="language-en" type="submit" value="" class='language-item btn'>
-                    {% trans "English" %}
-                    <span class="badge">5</span>
-                </button>
+                {% for key, value in LANGUAGES %}
+                    <button name="language-{{ key }}" value="{{ key }}" class='language-item btn'>
+                        {{ value }}
+                        <span class="badge">{{ trans_fields|get_value_from_dict:key}}</span>
```
suutari-ai: The badge is supposed to show the number of missing translations as far as I understand. So it should be updated by Javascript when new translations are entered.

E.g. if the form has following data:
```
name_en: Test
name_fi: Testi
name_sv:

description_en: This is a test resource.
description_fi:
description_sv:

specific_terms_en:
specific_terms_fi:
specific_terms_sv:
```
then the badges should show:
 * EN: 0  (no missing translations)
 * FI: 1  (missing = description)
 * SV: 2 (missing = name + description)

The "specific_terms" field is not counted as missing, since it doesn't have value in any language.

respa_admin/static_src/js/resourceFormLanguage.js
```diff
@@ -0,0 +1,98 @@
+/*
+* Toggle language based on entered language for all fields in the
+* whole form.
+*
+* @param {String} language  Language provided by the DOM value.
+* */
+export function toggleLanguage(language) {
+  if (language === getCurrentLanguage()) { return; }
```
suutari-ai: Why current language cannot be toggled?

respa_admin/templates/_base.html
```diff
@@ -15,6 +18,12 @@
 </head>
 
 <body>
+    <div class="hidden" id="all-languages">
+        {% for key, value in LANGUAGES %}
+            <input value="{{ key }}" />
```
suutari-ai: Maybe something else than `<input>` would be better for this, since it's not really an input, but just a way to send the languages data to the Javascript code. How about using `<span>` for this.?

suutari-ai:
> Check the unit tests. They seem to assume that name is a required field, but for some reason it's not anymore. Shouldn't one of the name fields (name_fi, name_en, or name_sv) be required though?

This seems to resolved now (with ff80ffb). Great! :+1:

respa_admin/static_src/js/resourceFormLanguage.js
dvainio: It is to prevent the user from hiding all of the translated fields from the UI.

suutari-ai: IMHO it's more harmful than useful. It makes the UI feel broken.

PR #349, opened 2018-09-19
Simple migration fixes

suutari-ai:
Update the migrations/models so that "makemigrations" doesn't generate
anything.  These are just meta data changes, so no need for new
migrations.


dvainio:
Everything OK.

PR #351, opened 2018-09-22
Update the message files

suutari-ai:
Update the message files with "manage.py makemessages" command and also
remove locations from them with "manage.py makemessages --no-location".


suutari-ai:
Self-merging as trivial.

suutari-ai:
@dvainio FYI: did this.

PR #357, opened 2018-09-25
respa_admin: Turn on username log-in by default

suutari-ai:
Turn on the log-in by username feature by default, because preserving
working Tunnistamo configuration in the Kontena deployment has had some
problems and therefore the username login is a better option for a while.


suutari-ai:
Merging this, since the change is trivial.

PR #361, opened 2018-09-26
[WIP] Implement sub resource and sub reservation support

tuomas777:
# !!! WIP DON'T MERGE !!!

## Sub resource and sub reservation support

This PR includes somewhat working POC implementation of sub resources and sub reservations. Because the changes needed are quite big regarding data model and API, and thus need to be right the first time (and there might be some better way to do this I'm missing), I present some analysis of the feature here.

### Motivation

We need a way to book a guide to a resource at the same time that the resource is booked. In the future there might be also other needs for linked resources, for example a resource that is divided to a number of smaller resources.

### Implemented solution

We create a way to link resources so that there can be parents and sub resources, and add the guides as sub resources. Then reservations also need to contain more than one resource, so we add sub reservations that contain info about the reserved sub resources.

*pros*
- pretty easy to implement
- existing code can be used for both sub resources and sub reservations
- API for sub resources and sub reservations will virtually the same as it is for parent resources and reservations (unless we want to remove unnecessary stuff)

*cons*
- there will be loads of unnecessary data stored for both sub resources and sub reservations
- the code will became a bit messier unless exhaustive refactoring is carried out

### Other considered solutions

* Use a separate model for guides (more work and probably too specific for future use).

* Add just guides / sub resources to reservations instead of using complete sub reservations. (much more work as most of reservation specific business logic need to be re-implemented and new API functionality added)


petterill:
[Review WIP PR 361 (Implement sub resource and sub reservation support)](https://trello.com/c/dfO0GWuK/2-review-wip-pr-361-implement-sub-resource-and-sub-reservation-support)

PR #363, opened 2018-09-27
Improve Django Admin UX for Users

suutari-ai:
Improve the user experience of managing User objects in Django Admin by
making the "General Administrator" field visible and more fields visible
in the user list view.  Also improve the Finnish transaltions of the
fields and their help texts.


suutari-ai:
The user list looks like this after this change:

image

and the permissions part on the User editing form:

image

PR #365, opened 2018-10-10
Respa Admin

suutari-ai:
This implements the Respa Admin user interface for managing the resources. It's a combination of work from @dvainio, @suutari-ai, @JsseL, @dotanthanh and @Djiffit.

There's some changes to the models in the `resources` and `users` apps too, including migrations, and the permission model has been changed according to the [new specification in `docs/permissions.rst`](https://github.com/City-of-Helsinki/respa/blob/0e3cbbff9e6ff3e1ddbe7c580c33dd9489499f08/docs/permissions.rst).

Note: There's a big risk of conflicts between this branch and master, because of the model/migration and permission changes, if merging of this is postponed for too long.


suutari-ai:
Btw. everything in this PR is already reviewed once, so there should be no need for a line by line reviews with a magnifying glass anymore. :mag:

petterill:
[Merge PR 365 (Respa admin)](https://trello.com/c/gimvtY3B/4-merge-pr-365-respa-admin)

JsseL:
There seems to be migrations with conflicting numbering e.g.:
- `resources/migrations/0067_add_reservation_confirmed_notification_extra.py`
- `resources/migrations/0067_resource_equipments_field.py`

In master branch the latest migration number under `resources/migrations/` seems to be `0067`. Se these migrations need to be regenerated to update the numbering.

This numbering conflict is problematic since Django determines the migration order from the autogenerated migration number. So the conflicting numbering needs to be solved before merging.

One way to fix this, is to remove the problematic migrations and rerun `manage.py makemigrations resources`.

suutari-ai:
> There seems to be migrations with conflicting numbering...

True, there are two migrations in the resources app with number 0067.

> So, the numbering needs to be fixed before merging.

No, it's already fixed. The conflict is resolved in `0068_merge_20180907_0708.py`.

You can also see that the migrations work fine, since currently there is no differences between the migrations in `respa-admin` branch and `master` and still the Docker image of `respa-admin` works fine.

> One way to fix this, is to remove the problematic migrations and rerun `manage.py makemigrations resources`.

That's one way, but the merge migration is also a good way to fix it and that's already done. Merge migration is also a better way in that sense that it won't invalidate any existing database (i.e. a database created from some earlier version of the `respa-admin` branch).

suutari-ai:
> currently there is no differences between the migrations in `respa-admin` branch and `master`

What I meant was that currently there is no changes to migrations in `master` branch since the last merge of `master` to `respa-admin` (i.e. Git commit range `respa-admin..master` doesn't contain any migrations).

JsseL:
@suutari-ai Great. Didn't look well enough into the migrations, but good that it's already taken care of. Sorry if I caused unnecessary head scratch. 😄

Issue #294, opened 2018-06-07
Implement linking resources


petterill:
Users wants to book a room and a service/personnel associated with that room, for example a music studio & guide person for that. These calendars should be "linked" in some way, so that Varaamo knows which resources can be booked simultaneously.

Some specs here: https://docs.google.com/presentation/d/1Gzksr89GCPLspS-3WGAsUbCG65wzkTn6kvAsEulzcX0/edit?usp=sharing

juyrjola:
Removing milestone because the UI changes will not be ready in time anyway.

PR #381, opened 2018-11-14
Child resource and child reservation support

tuomas777:
# Child resource and child reservation support

Related to Varaamo, one needs to be able to reserve also a guide when reserving a room. This need is tackled by implementing support for child resources and child reservations. This implementation is, at least currently, quite Varaamo specific.

## Child resources

Child resources are the same objects as normal resources. When a resource is added as at least one resource's child resource, it is considered a child resource. One resource can be a child resource for many normal resources, and one normal resource can have multiple child resources. Child resources cannot have child resources. Child resources are meant for the guides for now, but should be expandable for other uses as well later. 

New fields in resource API:

* list endpoint:
  * `child_resources`: IDs of child resources

* detail endpoint:
  * `child_resources`: all data of child resources nested
  * `parent_resources`: IDs of parent resources 

Child resources don't exist in list endpoint data as top level objects. 

## Child reservations

Child resources cannot be reserved on their own, only alongside a parent resource. In order to support child resource reserving, reservation objects are extended with child reservation support. Child reservations are basically normal reservation objects having `parent` field set to match their parent. A reservation can have multiple child reservations, but a child resource can have only one parent. Child reservations cannot have child reservations.

New fields in reservation API:

* list endpoint:
  * `child_reservations`: all data of child reservations nested

* detail endpoint:
  * `child_reservations`: all data of child reservations nested
  * `parent_reservation`: ID of parent reservation

Child reservations don't exist in list endpoint data as top level objects. 

Child reservations are created by nesting child reservations' data in `child_reservations` field in reservation data when POSTing a reservation. It is also possible (currently only for admins) to create a child reservation for an existing reservation by POSTing a single reservation and setting `parent_reservation` field.

Current rules regarding child reservations:

* Child reservation's begin and end times must be inside the parent's begin and end times

* Child reservation's begin time must match the parent's begin time

* Child reservations cannot be modified, only deleted and only by admin

* Child reservations can be added to existing reservations only by admins

* Child reservations don't support manual confirmation functionality, they are always either `CONFIRMED` or `CANCELLED`

In the future these might be made customizable to make the implementation more general usage suitable.

## Notifications

There are two new notification types related to child reservations:

* `CHILD_RESERVATION_CREATED_SEPARATELY`: sent when a child reservation is added to an existing reservation

* `CHILD_RESERVATION_CANCELLED`: sent when a child reservation is cancelled

Normal reservation notifications will also have related child reservations' data available in notification context `child_reservations` field.

## TODO

* Prevent creating of child resources for child resources

* Add documentation of the feature to the repo

* Update the Swagger document

* Check performance, there might be issues, currently includes only pretty trivial prefetching

Also, **this feature requires careful code review and TESTING (ofc tested locally and the PR includes "unit" tests, but the feature is quite a big change codewise and time basically ran out while implementing it, so it is possible something is still missing / broken)**

Refs #294


jussiarpalahti:
Closing as superseded by #feature/linked-resources

PR #391, opened 2018-11-16
Add route to location link to email notification

### LINKED ISSUE #338: Add route to location link to email notification
The notification email should include a link to Palvelukartta that automatically displays a link to the Unit of the Resource that was booked. Similar thing is available on resource page on Varaamo for example https://varaamo.hel.ninja/resources/au2tzfwtc34q?date=2018-09-07 -> https://palvelukartta.hel.fi/unit/8215#!route-details. 

This requires that the Unit ID can be referenced in the email body. Currently only the name of the Unit can be referenced. The actual link to Palvelukartta will be constructed with html.

jussih:
Closes #338 

Unit id added to Notification template context as `unit_id`.

### DIFF
--- a/resources/models/reservation.py
+++ b/resources/models/reservation.py
@@ -343,6 +343,7 @@ class Reservation(ModifiableModel):
             }
             if self.resource.unit:
                 context['unit'] = self.resource.unit.name
+                context['unit_id'] = self.resource.unit.id
             if self.can_view_access_code(user) and self.access_code:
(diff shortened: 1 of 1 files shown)


jussih:
@petterill The original notification message templates have been added to the database with a data migration. Shall we create a new data migration to update those, or will the template configuration be done by hand in production to include the new links?

If we should create a data migration, we need a wording or guidance for the templates from the product owner for the desired user experience.

PR #394, opened 2018-11-21
Implement support for "detailed instructions"

### LINKED ISSUE #378: Implement support for "detailed instructions"
Support for instructions that describe how to use and access the resource. A new field that can be displayed in Varaamo and support for bringing this info to email notifications.

"Tilan käyttöön liittyvät erityiset ohjeet tarvittaessa, tähän tieto myös opastavan henkilökunnan läsnäolosta"

jussih:
Closes #378

Adds new translated text attribute to Resource for detailed reservation instructions.

The attribute is `Resource.reservation_instructions`, it is exposed in respa_admin and visible in the API:

```
GET /v1/resource/avq63jqdhleq/

{
    "id": "avq63jqdhleq",
...
    "reservation_instructions": {
        "en": "Foo"
    },
}
```


resources/models/resource.py
```diff
@@ -187,6 +187,9 @@ class Resource(ModifiableModel, AutoIdentifiedModel):
                                                     null=True, blank=True)
     reservable = models.BooleanField(verbose_name=_('Reservable'), default=False)
     reservation_info = models.TextField(verbose_name=_('Reservation info'), null=True, blank=True)
```
juyrjola: We talked about this with @petterill, and figured out the difference between this field and the new one. Please add a `help_text` attribute to `reservation_info` with the contents "If there are any special considerations for making reservations, write them here".

resources/models/resource.py
```diff
@@ -187,6 +187,9 @@ class Resource(ModifiableModel, AutoIdentifiedModel):
...
+    reservation_instructions = models.TextField(
```
juyrjola: This should be renamed to `usage_instructions`. `verbose_name` should be just `_('Usage instructions')` and `help_text` should be "Optional instructions regarding the use of the resource, including required guide presence".

resources/models/resource.py
jussih: This change done in 8affc1d.

resources/models/resource.py
jussih: This change done in 8affc1d.

juyrjola:
Hmm, should `respa_admin/forms.py` be also modified for the field to be visible in the admin UI?

juyrjola:
I'm not too familiar with the admin UI yet...

petterill:
It should be possible to input these through Respa Admin yes, if that's what @juyrjola means. Is this implemented already in this @jussih ?

PR #397, opened 2018-11-30
Feature/add support for ical

otiala:
Add support for ical files utilizing the existing function to return valid ical-formatted string.


resources/models/reservation.py
```diff
@@ -24,6 +24,7 @@
     get_dt, save_dt, is_valid_time_slot, humanize_duration, send_respa_mail,
     DEFAULT_LANG, localize_datetime, format_dt_range
 )
+from resources.models.utils import build_reservations_ical_file
```
jussih: This could be among the other `.utils` imports above.

resources/models/reservation.py
```diff
@@ -419,16 +421,28 @@ def send_reservation_denied_mail(self):
         self.send_reservation_mail(NotificationType.RESERVATION_DENIED)
 
     def send_reservation_confirmed_mail(self):
-        self.send_reservation_mail(NotificationType.RESERVATION_CONFIRMED)
+        reservations = [self]
+        ical_file = build_reservations_ical_file(reservations)
+        attachment = 'reservation.ics', ical_file, 'text/calendar'
```
jussih: This is a tuple, I would use parenthesis to make it clear
`('reservation.ics', ical_file, 'text/calendar')`

resources/models/utils.py
```diff
@@ -108,10 +110,12 @@ def send_respa_mail(email_address, subject, body, html_body=None):
 
     notification_logger.info('Sending notification email to %s: "%s"' % (email_address, subject))
 
+    subject, from_email, to = subject, from_address, email_address
+    text_content = body
+    msg = EmailMultiAlternatives(subject, text_content, from_email, [to], attachments=attachments)
```
jussih: Why not just use the original variables as parameters to EmailMultiAlternatives instead of mapping them to new names?

resources/models/reservation.py
otiala: Good catch, done.

resources/models/reservation.py
otiala: Done.

resources/models/utils.py
otiala: I assume this was a remain from refactoring the `send_respa_mail()` method. Changed.

PR #398, opened 2018-12-04
Created missing migrations for changes in resource and unit models

turtommy:


turtommy:
I noticed that makemigrations tries to create migration file that changes some fields in unit and resource models.

PR #416, opened 2019-01-03
More configuration from environment, use TOKEN_AUTH_* for authentication

vikoivun:


respa/settings.py
```diff
@@ -17,8 +17,9 @@
     ALLOWED_HOSTS=(list, []),
     ADMINS=(list, []),
     DATABASE_URL=(str, 'postgis:///respa'),
-    JWT_SECRET_KEY=(str, ''),
-    JWT_AUDIENCE=(str, ''),
+    SECURE_PROXY_SSL_HEADER=(tuple, None)
```
tuomas777: Missing a comma at the end.

jussiarpalahti:
@vikoivun So, replacing variables are same as replaced variables, only names are changed? No change in expected values?

vikoivun:
> @vikoivun So, replacing variables are same as replaced variables, only names are changed? No change in expected values?

Yep. It just tries to show in the name, that is has to do with tokens and authentication.

vikoivun:
The sentry logging thing is a weird one. At one point at least we had a case where anymail had just silently eaten any errors with sending mail. But recently when mailgun reached the free message limit, anymail instead raised the exception causing the whole reservation to fail.

jussiarpalahti:
What sentry logging thing? I see there's both raven and sentry error loggers in config now. Are both really necessary?

vikoivun:
> What sentry logging thing? I see there's both raven and sentry error loggers in config now. Are both really necessary?

As I said, at some point I was certain that anymail logged errors and I wanted to capture them through sentry. But now I looked through anymail code and there are no loggers at all. I'll remove the logging.

jussiarpalahti:
Well, yes, but there is configs for raven (the old sentry) and sentry.errors which I take is for new SDK. Can both truly work with either Python package?

PR #420, opened 2019-01-11
Add municipality filter to ResourceFilterSet

### LINKED ISSUE #311: Implement searching/filtering resources according to City (Espoo/Helsinki/Vantaa)
There is a municipality field in Respa, but itäs is not possible to filter API calls by municipality

otiala:
Closes #311

### DIFF
--- a/resources/api/resource.py
+++ b/resources/api/resource.py
@@ -323,6 +323,8 @@ class ResourceFilterSet(django_filters.FilterSet):
                                       widget=django_filters.widgets.CSVWidget, distinct=True)
     available_between = django_filters.Filter(method='filter_available_between',
                                               widget=django_filters.widgets.CSVWidget)
+    municipality = django_filters.Filter(name='unit__municipality', lookup_expr='in',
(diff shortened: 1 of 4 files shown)


resources/api/resource.py
```diff
@@ -323,6 +323,8 @@ def __init__(self, *args, **kwargs):
                                       widget=django_filters.widgets.CSVWidget, distinct=True)
     available_between = django_filters.Filter(method='filter_available_between',
                                               widget=django_filters.widgets.CSVWidget)
+    municipality = django_filters.Filter(name='unit__municipality', lookup_expr='in',
```
jussih: By what attribute of Municipality is the filtering done by? Other filters here specify the attribute such as `equipment = django_filters.Filter(name='resource_equipment__equipment__id',`.

To be specific it would be good to define the attribute name here.

otiala: Filtering is done by id atrribute, which is currently the name of municipality in all lower case letters, such as: `hämeenlinna` (including scandic letters). I agree with being clear with the filtering attribute, I will change it.

PR #421, opened 2019-01-14
Add onclick remove to notifications

### LINKED ISSUE #409: Fix "sign in succesful" notification
Kirjauduttiin sisään käyttäjänä xxx should close when clicked on x (close). It does not currently close.

otiala:
Closes #409 

Add onclick remove to notifications

### DIFF
--- a/respa_admin/static_src/js/base.js
+++ b/respa_admin/static_src/js/base.js
@@ -1,0 +1,19 @@
+// Base events that will be loaded in base.html
+
+/*
+* Bind the event handler for closing notification to the elements (buttons)
+* since we cant do it directly in the html because of the scope
+* */
+function enableNotificationHandler() {
+    let notifications = document.getElementsByClassName('noti');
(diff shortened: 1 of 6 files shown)


otiala:
> There seems to be some code already in https://github.com/City-of-Helsinki/respa/blob/master/respa_admin/static_src/js/resourceForm.js#L57 for closing the notifications. Could you take another look and see if that code could handle this case instead of the inline js snippet?

Good catch! However, the notification handler probably shouldn't be in `resourceForm.js`, as it is initialized only in `create_resource.html`. I moved the notification handler to a new file called `base.js`, which will be initialized in `base.html`, to avoid loading unnecessary event listeners.

otiala:
@jussiarpalahti @petterill as a side note, this should be tested in test server before releasing to production, as I had no proper way to test it locally, due to the requirement of using Helsinki account to log in.

PR #427, opened 2019-01-22
Implement Kulkunen access control system

juyrjola:
This PR implements support for controlling external access control systems in Respa. The system can support different ACS drivers, but only Siemens SiPass is implemented for now. Please see the included [documentation](https://github.com/City-of-Helsinki/respa/blob/3fe3ff5167587de606affe2a8c989f6461860c55/docs/kulkunen.md) for more information.

There are no unit tests implemented yet as agreed with the PO, but they should be done before Kulkunen undergoes any further major development.


kulkunen/models.py
```diff
@@ -0,0 +1,387 @@
...
+    class Meta:
+        index_together = (('resource', 'state'),)
+
+    def __str__(self) -> str:
+        # time_range = format_dt_range(settings.LANGUAGES[0][0], self.starts_at, self.ends_at)
+        return _("{user} {reservation} ({state})").format(
+            user=self.user, reservation=self.reservation, state=self.state
+        )
+
+    def is_active(self):
```
jussih: `AccessControlGrantQuerySet.active()` has different rules compared to `AccessControlGrant.is_active()`. If they intentionally do different things, then the naming is confusing and could be changed to differentiate the results.

kulkunen/models.py
```diff
@@ -0,0 +1,387 @@
...
+        if driver_class is None:
+            for name, verbose_name, driver_path in DRIVERS:
+                if name == self.driver:
+                    break
+            else:
+                raise ImproperlyConfigured("Driver %s not found" % self.driver)
+
+            parts = driver_path.split('.')
+            module_path, class_name = '.'.join(parts[:-1]), parts[-1]
+            module = import_module(module_path)
```
jussih: Django also has dotted string import as a prepackaged util: https://github.com/django/django/blob/master/django/utils/module_loading.py#L7

kulkunen/drivers/sipass.py
```diff
@@ -0,0 +1,682 @@
...
+    expires_at: datetime
+
+    def __init__(self, value, expires_at):
+        self.value = value
+        self.expires_at = expires_at
+
+    def has_expired(self):
+        if self.expires_at is None:
+            return False
+        now = datetime.now()
```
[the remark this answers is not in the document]
juyrjola: This one I left as it is, because the datetimes are serialized as Unix timestamps and only used internally by the SiPass driver code.

juyrjola:
Thanks for the review @jussih! I've now pushed a version where the review comments are taken into account and the commits are rebased against current `master`.

PR #428, opened 2019-01-24
Fix/407 filter paid resources

### LINKED ISSUE #407: Fix filtering paid/unpaid resources
Filtering for paid/unpaid resources does not work:

Ville:

Tällä hetkellä varaamon haussa voi tarkennetun haun puolella asettaa täppä et näytä vain maksuttomat tilat. Tämä näemmä laittaa `charge` url parametrin trueksi respakutsuun (esim. https://respa.koe.hel.ninja/v1/resource/?charge=true&end=2019-01-09T23%3A59%3A59Z&page=1&start=2019-01-09T00%3A00%3A00Z&page_size=30), mutta tällä hetkellä tämä ei aiheuta tuloksen filtteröintiä.

Jos tämä filtteri speksattaisiin niin et respa filtteröi täpän ollessa päällä tulosjoukkoa niin et näytettäisiin vain ne jotka ovat maksuttomia tai sellaiset joiden minimihinta lähtee 0:sta, niin millä aikataululla onnistuis tällasen toteutus?

Mietin kans et jos kerran toi parametri ei vielä toimi niin pitäiskö sen nimeä vielä miettiä uudelleen? Tällä hetkellä jos “Maksuton tila” valinta on päällä niin parametrin `charge` arvo on `true`. Jos default tilassa (`charge`: `false`) on tarkoitus näyttää kaikki ja sitten arvolla `true` näyttää vain maksuttomat niin olisiko parempi nimi parametrille esim. `only_free_of_charge` tai `free_of_charge` (edited)

otiala:
Closes #407 

Implemented a boolean filter `free_of_charge` for `/resource` endpoint.

`free_of_charge=true` returns resources that have their `min_price_per_hour` value of `0` or `None`, as in not set.

`free_of_charge=false` returns all resources, as this is the default parameter and value in Varaamo requests.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -460,6 +460,11 @@ paths:
           required: false
           type: number
           description: Use together with `lat` and `long`. Returns only resources that are within the given `distance` of the point which is calculated from `lat` and `lon`.
+        - name: free_of_charge
+          in: query
+          required: false
+          type: boolean
+          description: If given boolean is `true`, returns only resources that have their `min_price_per_hour` value of `0` or `None`. If given boolean is `false`, returns all resources.  
       responses:
         '200':
           description: Successful response
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -464,7 +464,7 @@ paths:
           in: query
           required: false
           type: boolean
-          description: If given boolean is `true`, returns only resources that have their `min_price_per_hour` value of `0` or `None`. If given boolean is `false`, returns all resources.  
+          description: If given boolean is `true`, returns only resources that have their `min_price_per_hour` value of `0` or `None`. If given boolean is `false`, returns resources that have their `min_price_per_hour` greater than 0.  
       responses:
         '200':
           description: Successful response
--- a/resources/api/resource.py
+++ b/resources/api/resource.py
@@ -323,6 +323,8 @@ class ResourceFilterSet(django_filters.FilterSet):
                                       widget=django_filters.widgets.CSVWidget, distinct=True)
     available_between = django_filters.Filter(method='filter_available_between',
                                               widget=django_filters.widgets.CSVWidget)
+    free_of_charge = django_filters.BooleanFilter(method='filter_free_of_charge',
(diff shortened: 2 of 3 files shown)


swagger.yaml
```diff
@@ -460,6 +460,11 @@ paths:
           required: false
           type: number
           description: Use together with `lat` and `long`. Returns only resources that are within the given `distance` of the point which is calculated from `lat` and `lon`.
+        - name: free_of_charge
+          in: query
+          required: false
+          type: boolean
+          description: If given boolean is `true`, returns only resources that have their `min_price_per_hour` value of `0` or `None`. If given boolean is `false`, returns all resources.
```
juyrjola: Maybe this was already discussed, but shouldn't the logical behavior be to return _only paid_ resources if `free_of_charge` is set to `false`? If the API user doesn't want to filter based on the charge status, they can leave out the filter altogether.

otiala: @juyrjola as far as I know, by default respa sends the parameter as `false` on all requests when fetching the resources, so if we implement the filter as suggested, Varaamo needs some refactoring. Another solution could be to rename the filter to `only_free_of_charge`, as suggested by @vikoivun (I think?) on the original issue #407, so it would make more sense.

juyrjola: Currently Varaamo (as deployed at varaamo.hel.fi) appears to use a query param called `charge`, which is unsupported anyway. Looking at the Varaamo code, changing it to only send the query param if the filter is active looked like a pretty small change (it can be done e.g. here: https://github.com/City-of-Helsinki/varaamo/blob/feature/helsinki-brand-v2/app/utils/searchUtils.js#L8)

juyrjola: So something like this, and a small change in Varaamo → voilà!
 
```python
    def filter_free_of_charge(self, queryset, name, value):
        qs = Q(min_price_per_hour__lte=0) | Q(min_price_per_hour__isnull=True)
        if value:
            return queryset.filter(qs)
        else:
            return queryset.exclude(qs)
```

PR #438, opened 2019-01-31
Change timeslots

### LINKED ISSUE #434: Fix max reservable time dropdown
Add half hour time slots between full hour time slots. Time slots should be 00:30 to 23:30 and the dropdown should have a slider so that is does not fill the whole screen. See how this is implemented in Varaamo search.

otiala:
Closes #434 

Changed the timeslots of `max_period` to be from 00:30 to 23:30, with increments of 00:30.

Removed 00:45 timeslots.

@petterill changing behaviour of HTML dropdown elements is not possible with CSS, as the size of the list and the scrollbar-element is decided by browser. I added a picture of how it looks atm.

If this is not okay, we can take a hacky JS-approach, or think of something else than a dropdown.

dropdown


jussih:
Looks like this on my machine. I think it's fine - it's usually best to let the browser render the selector natively. It'll probably work better on mobile phones as well compared to a fully custom javascript component.

screenshot from 2019-02-04 15-38-36

PR #440, opened 2019-01-31
Use tprek v2 API

### LINKED ISSUE #372: Switch to using Servicemap API v2 instead of v1
Servicemap API is used as a source for Unit ID's and related info. V1 doesn't update properly anymore and will stop functioning in the near future.

Let's also consider taking care of  #169 while doing this.

jussih:
Closes #372 

Note 1: Servicemap API currently returns 500 when filtering with more than 1 service. This should be fixed in the servicemap service.

Note 2: API v2 has translated data in the `address_postal_full` field, whereas v1 API had no data in the field. Currently the postal address is not saved as the field is not translatable in respa. We could enable translations with a schema migration and save the data.

### DIFF
--- a/resources/importer/tprek.py
+++ b/resources/importer/tprek.py
@@ -40,7 +40,9 @@ class TPRekImporter(Importer):
             point = Point(x=coords[0], y=coords[1], srid=4326)
             data['location'] = point
 
-        data['modified_at'] = dateutil.parser.parse(data['origin_last_modified_time'])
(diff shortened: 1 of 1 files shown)


tituomin:
@jussih The Servicemap API has been updated and the problem has been fixed. (Note 1)

jussih:
> @jussih The Servicemap API has been updated and the problem has been fixed. (Note 1)

I tested the importer and it works now :+1: 

Regarding Note 2, it was decided not to use the `address_postal_full` data at this point. This is now ready for merging.

PR #457, opened 2019-02-07
Implement order by functionality

### LINKED ISSUE #449: Order resource listing by name, visibility & availability
These are clickable elements at the top of the resource listing but clicking on them does nothing. The functionality has not been implemented.

otiala:
Closes #449 

Implemented the functionality to order resources by `name`, `availability` or `visibility`.

Clicking one of the mentioned buttons will redirect to the same page, setting GET parameters (t.ex. `?order_by=name`), and clicking it again will cause the parameter value to be prefixed with `-`.

The GET parameter value is used directly in the `order_by()` method of the queryset, and exceptions will return an unfiltered queryset.

### DIFF
--- a/respa_admin/views/resources.py
+++ b/respa_admin/views/resources.py
@@ -65,8 +65,9 @@ class ResourceListView(ListView):
         if self.resource_unit:
             qs = qs.filter(unit=self.resource_unit)
         if self.order_by:
+            order_by_param = self.order_by.strip('-')
             try:
-                if Resource._meta.get_field(self.order_by):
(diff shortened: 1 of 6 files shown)


respa_admin/views/resources.py
```diff
@@ -61,6 +64,12 @@ def get_queryset(self):
             qs = qs.filter(type=self.resource_type)
         if self.resource_unit:
             qs = qs.filter(unit=self.resource_unit)
+        if self.order_by:
```
jussih: There could be a simple test case which runs the order_by code here. Currently it's not covered by any test.

PR #461, opened 2019-02-11
Alter resource fields

### LINKED ISSUE #454: Disable negative number input
It is possible to input negative numbers to all fields that require numbers. Some fields check for this, some don't. At least the numbers of persons and surface area fields accept negative numbers and do not produce an error message. Check all fields and either disallow adding negative numbers so that they are not an option to select or add error messages and checks to all fields.

otiala:
Closes #454 

Changed fields `area`, `max_reservations_per_user` and `people_capacity` to `PositiveIntegerField` for automatic validation.


resources/migrations/0074_alter_resource_fields.py
```diff
@@ -0,0 +1,30 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('resources', '0073_add_resource_generate_access_codes_field'),
+    ]
+
+    operations = [
```
jussih: Have you tested what happens if the database contains negative numbers in these fields when executing the migration?

To be safe you could add an operation here which is executed first and sets all negative values to 0 in these fields. Then the alter columns will surely succeed.

PR #464, opened 2019-02-13
Days in opening periods sorted properly in Respa Admin

### LINKED ISSUE #451: Fix displaying of days in opening hours section
When one creates a new time slot for opening hours (for example 5.2. - 8.2. tuesday to friday), the names of days are filled in correctly. However if one changes the dates (to for example 6.2. - 9.2. wednesday to saturday) the names of days remain the same. These should be updated to match the new dates correctly. The correct behaviour would be to always have the first day of the period first. If for example the period is 6.2. - 2.8.2., then the list would begin at tuesday and go through all the dates so that monday would be the last. This way it's probably less confusing to the user.

Also I guess it should not be possible to change the name of the date. For example one could change tuesday's name to wednesday. Is there any logic behind this?

jussih:
Closes #451 

The days are always sorted starting from the weekday that is selected as the starting day.
Weekday names are no clickable, so the user can not select for example two Tuesdays, but has to obey the weekdays generated via selecting the start and end days.

### DIFF
--- a/respa_admin/static_src/js/resourceFormPeriods.js
+++ b/respa_admin/static_src/js/resourceFormPeriods.js
@@ -110,13 +110,14 @@ function updatePeriodDaysIndices($periodItem) {
     })
   };
 
-  let index = 0;
-  for (;index < originalDaysList.length; index++) {
-    setIndex(index, originalDaysList[index]);
+  let dayIndex = 0;
+  for (let i = 0;i < originalDaysList.length; i++) {
+    setIndex(dayIndex, originalDaysList[i]);
+    dayIndex++;
   }
-  for (let newDayLoopIndex = 0;newDayLoopIndex < newDaysList.length; newDayLoopIndex++) {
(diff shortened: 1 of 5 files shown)


respa_admin/static_src/js/resourceFormPeriods.js
```diff
@@ -78,7 +108,16 @@ function updatePeriodDaysIndices($periodItem) {
         $(cellInput).attr('name', $(cellInput).attr('name').replace(/-(\d+)-(\d+)-/, '-' + periodIdNum + '-' + dayIndex + '-'));
       });
     })
-  });
+  };
+
+  let index = 0;
```
otiala: Maybe a comment on why the `index` variable is defined outside the for-loop would be good for someone reading the code in the future, it confused me at first.

jussih: This was indeed unlegible code. It's clearer now.

PR #467, opened 2019-02-14
Disable access code generation for resources that are connected to the SiPass Kulkunen driver

juyrjola:


kulkunen/drivers/sipass.py
```diff
@@ -682,3 +682,19 @@ def prepare_install_grant(self, grant):
         # every nightly.
         grant.install_at = grant.starts_at - timedelta(days=1)
         grant.save(update_fields=['install_at'])
+
+    def save_respa_resource(self, resource, respa_resource):
+        # SiPass driver generates access codes by itself, so we need to
+        # make sure Respa doesn't generate them.
+        if not respa_resource.generate_access_codes:
+            return
+        respa_resource.generate_access_codes = False
```
jussih: `save_respa_resource` and `save_resource` implementations here are confusing me. It seems to me these two methods are having a temporal dependence - the results are different based in what order the methods are called. If this one is executed first, the Resource is probably never saved as the other bails out before saving (if it's the same instance of Resource).

Maybe there's just a `respa_resource.save()` call missing here though?

jussih: I see, this is called from a `pre_save` signal handler, which explains the missing `save()` call here.

As both these methods and the signal handler are part of the higher level Kulkunen abstraction, could the method be named differently to reflect that the implementation doesn't actually need to do any saving?

juyrjola: Yes, the other is called when a Respa resource is being saved and the other when an access control resource is saved. Either action might cause need for the ACS driver to intervene. I considered naming them `notify_*`, but then that might be confused with `notify_access_code()`, which is used for sending a notification message to the user. Another option would be `process_respa_resource_save`, but that's already a lot of words. What the methods are supposed to do is _indicate to the ACS driver that an object is being saved, so that the driver can perform the modifications it needs on the object itself or a corresponding other object_.

I did notice that the docstrings for the methods were written a bit too "löysin rantein", so I updated them to better match reality. Perhaps that is enough to make it clearer? Would you have some other naming suggestions?

jussih: The comment makes it clear now for the driver implementer.

PR #477, opened 2019-02-18
Make private resources visible for managers and admins

### LINKED ISSUE #462: Fix Unit Administrator not seeing hidden resources

otiala:
Closes #462

### DIFF
--- a/resources/auth.py
+++ b/resources/auth.py
@@ -23,3 +23,8 @@ def is_unit_admin(user):
     return is_authenticated_user(user) and (
         user.unit_group_authorizations.admin_level().exists() or
         user.unit_authorizations.admin_level().exists())
+
+
+def is_unit_manager(user):
+    return is_authenticated_user(user) and (
+        user.unit_authorizations.manager_level().exists())
--- a/resources/auth.py
+++ b/resources/auth.py
@@ -17,3 +17,9 @@ def is_any_admin(user):
         is_general_admin(user) or
         user.unit_group_authorizations.admin_level().exists() or
         user.unit_authorizations.admin_level().exists())
+
+
+def is_unit_admin(user):
+    return is_authenticated_user(user) and (
+        user.unit_group_authorizations.admin_level().exists() or
+        user.unit_authorizations.admin_level().exists())
--- a/resources/auth.py
+++ b/resources/auth.py
@@ -17,14 +17,3 @@ def is_any_admin(user):
         is_general_admin(user) or
         user.unit_group_authorizations.admin_level().exists() or
         user.unit_authorizations.admin_level().exists())
-
-
-def is_unit_admin(user):
-    return is_authenticated_user(user) and (
-        user.unit_group_authorizations.admin_level().exists() or
-        user.unit_authorizations.admin_level().exists())
-
-
-def is_unit_manager(user):
-    return is_authenticated_user(user) and (
-        user.unit_authorizations.manager_level().exists())
--- a/resources/models/resource.py
+++ b/resources/models/resource.py
@@ -29,7 +29,7 @@ from PIL import Image
 from guardian.shortcuts import get_objects_for_user, get_users_with_perms
 from guardian.core import ObjectPermissionChecker
 
-from ..auth import is_authenticated_user, is_general_admin, is_unit_admin
+from ..auth import is_authenticated_user, is_general_admin, is_unit_admin, is_unit_manager
 from ..errors import InvalidImage
 from ..fields import EquipmentField
 from .base import AutoIdentifiedModel, NameIdentifiedModel, ModifiableModel
@@ -137,6 +137,11 @@ class ResourceQuerySet(models.QuerySet):
             administered_resources_q = Q(unit__in=administered_unit_ids)
             public_resources_q = Q(public=True)
             return self.filter(administered_resources_q | public_resources_q)
+        elif is_unit_manager(user):
+            managed_unit_ids = Unit.objects.managed_by(user).values_list('id', flat=True)
+            managed_resources_q = Q(unit__in=managed_unit_ids)
+            public_resources_q = Q(public=True)
+            return self.filter(managed_resources_q | public_resources_q)
         else:
             return self.filter(public=True)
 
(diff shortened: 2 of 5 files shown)


resources/models/resource.py
```diff
@@ -132,6 +132,16 @@ class ResourceQuerySet(models.QuerySet):
...
+        elif is_unit_admin(user):
+            administered_unit_ids = Unit.objects.administered_by(user).values_list('id', flat=True)
+            administered_resources_q = Q(unit__in=administered_unit_ids)
+            public_resources_q = Q(public=True)
+            return self.filter(administered_resources_q | public_resources_q)
+        elif is_unit_manager(user):
+            managed_unit_ids = Unit.objects.managed_by(user).values_list('id', flat=True)
+            managed_resources_q = Q(unit__in=managed_unit_ids)
+            public_resources_q = Q(public=True)
+            return self.filter(managed_resources_q | public_resources_q)
```
suutari-ai: Shouldn't these both cases be combined, i.e. if the user is admin for some units and manager for some others?

I don't think it's useful to do the `is_unit_admin` or `is_unit_manager` calls, since they will only cause unneeded SQL queries.

Also, it's not good to pull the IDs to Python, but rather do the query with SQL as a whole.

I would do it like this (untested though):
```python
def visible_for(self, user):
    if is_general_admin(user):
        return self
    in_admined_units = Q(unit__in=Unit.objects.administered_by(user))
    in_managed_units = Q(unit__in=Unit.objects.managed_by(user))
    is_public = Q(public=True)
    return self.filter(in_admined_units | in_managed_units | is_public)
```

suutari-ai: hmm.. but now that I look again at the code of `UnitQuerySet.managed_by`, I see that (general or unit) administrators are already considered there, so it should be enough to do it like this (without even adding the `administered_by` method to the UnitQuerySet):
```python
def visible_for(self, user):
    if is_general_admin(user):
        return self
    in_managed_units = Q(unit__in=Unit.objects.managed_by(user))
    is_public = Q(public=True)
    return self.filter(in_managed_units | is_public)
```

resources/models/unit.py
```diff
@@ -33,6 +33,22 @@ def managed_by(self, user):
...
+            return self
+
+        unit_group_q = Q(
+            unit_groups__authorizations__in=(
+                user.unit_group_authorizations.admin_level()))
+        unit_q = Q(
+            authorizations__in=(
+                user.unit_authorizations.admin_level()))
+
+        return self.filter(unit_group_q | unit_q).distinct()
```
suutari-ai: This new method should not be needed, since managed_by already considers the admins as users that can manage the units.

petterill:
@otiala @jussih please review the comments by Tuomas and make changes where needed

resources/models/unit.py
otiala: @suutari-ai Good cathes, the specification changed at some point, which apparently caused some confusion. However, I have changed the code now.

resources/auth.py
```diff
@@ -17,3 +17,14 @@ def is_any_admin(user):
         is_general_admin(user) or
         user.unit_group_authorizations.admin_level().exists() or
         user.unit_authorizations.admin_level().exists())
+
+
+def is_unit_admin(user):
```
jussih: It seems these are not used anywhere (`is_unit_admin` and `is_unit_manager`) after refactorings. Remove the dead code.

resources/models/resource.py
```diff
@@ -29,7 +29,7 @@
 from guardian.shortcuts import get_objects_for_user, get_users_with_perms
 from guardian.core import ObjectPermissionChecker
 
-from ..auth import is_authenticated_user, is_general_admin
+from ..auth import is_authenticated_user, is_general_admin, is_unit_admin, is_unit_manager
```
jussih: Seems `is_unit_admin` and `is_unit_manager` are not used. Remove imports.

resources/tests/test_resource_api.py
```diff
@@ -136,11 +137,41 @@ def test_non_public_resource_visibility(api_client, resource_in_unit, user):
...
+    user.save()
+    url = reverse('resource-list')
+    response = api_client.get(url)
+    assert response.status_code == 200
+    assert response.data['count'] == 1
+    assert Unit.objects.managed_by(user).values_list('id', flat=True)[0] == response.data['results'][0]['unit']
+
+    # Authenticated as unit group admin
+    user.unit_authorizations.level = UnitGroupAuthorizationLevel.admin
+    user.unit_authorizations.subject = resource_in_unit.unit
```
suutari-ai: I think you should clear the unit authorizations here and instead add a new unit *group* authorization.

resources/tests/test_resource_api.py
```diff
@@ -136,11 +137,41 @@ def test_non_public_resource_visibility(api_client, resource_in_unit, user):
...
-
     url = reverse('resource-detail', kwargs={'pk': resource_in_unit.pk})
     response = api_client.get(url)
     assert response.status_code == 200
 
+    # Authenticated as unit admin
+    user.unit_authorizations.authorized = staff_user
+    user.unit_authorizations.level = UnitAuthorizationLevel.admin
+    user.unit_authorizations.subject = resource_in_unit.unit
+    user.save()
```
suutari-ai: Hmm.. didn't notice that first, but this isn't adding a new authorization at all. Only reason it "works" is that the user is already a general admin. It doesn't verify anything though. So you should clear the `is_general_admin` flag and use the related manager to create the authorization:
```python
user.is_general_admin = False
user.save()
user.unit_authorizations.create(
    authorized=staff_user,
    level=UnitAuthorizationLevel.admin,
    subject=resource_in_unit.unit)
```

resources/tests/test_resource_api.py
```diff
@@ -136,11 +137,51 @@ def test_non_public_resource_visibility(api_client, resource_in_unit, user):
...
+    url = reverse('resource-list')
+    response = api_client.get(url)
+    assert response.status_code == 200
+    assert response.data['count'] == 1
+    assert Unit.objects.managed_by(user).values_list('id', flat=True)[0] == response.data['results'][0]['unit']
+
+    # Authenticated as unit manager
+    user.unit_authorizations.level = UnitAuthorizationLevel.manager
+    user.unit_authorizations.subject = resource_in_unit.unit
+    user.save()
```
suutari-ai: This is still incorrect and relying to the UnitGroup Administrator status set on line 163. Maybe you could move this as the first one of these three new tests and then in the second test just change the level of the UnitAuthorization object from manager to admin.

Otherwise these tests look fine now.

PR #491, opened 2019-02-26
Add functionality to display respa admin instructions

### LINKED ISSUE #488: Add a link to instructions
Before we add proper instructional texts, let's add a link to the Google Drive instructions file that have been written for Respa Admin. This is not ideal obviously but it helps for now and the instructions can be easily maintained. The link location should preferably be configurable.

This is how it was done in Kerrokantasi as well. Although there it seems to have become a permanent solution...

The proper location would probably be in the top menu, left of the user name. Add a question mark icon and the text Ohjeet / Instructions.

otiala:
Closes #488 

Implemented functionality to display a button in the UI to redirect to respa admin instructions.
It reads `RESPA_ADMIN_INSTRUCTIONS_URL` variable from `.env` file, and `ExtraContextMixin` mixin sets it into context, which is inherited in needed views. 

`RESPA_ADMIN_INSTRUCTIONS_URL` defaults to an empty string, and in that case the new Instructions-button wont be displayed in UI, so it won't cause any problems in case the said variable is not set.

### DIFF
--- a/respa/settings.py
+++ b/respa/settings.py
@@ -160,7 +160,6 @@ TEMPLATES = [
                 'django.template.context_processors.request',
                 'django.contrib.auth.context_processors.auth',
                 'django.contrib.messages.context_processors.messages',
-                'respa_admin.context_processors.export_global_vars',
             ],
         },
     },
--- a/respa/settings.py
+++ b/respa/settings.py
@@ -37,6 +37,7 @@ env = environ.Env(
     ACCESSIBILITY_API_BASE_URL=(str, 'https://asiointi.hel.fi/kapaesteettomyys/'),
     ACCESSIBILITY_API_SYSTEM_ID=(str, ''),
     ACCESSIBILITY_API_SECRET=(str, ''),
+    RESPA_ADMIN_INSTRUCTIONS_URL=(str, ''),
 )
 environ.Env.read_env()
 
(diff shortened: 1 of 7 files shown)


respa/settings.py
```diff
@@ -156,6 +157,7 @@
                 'django.template.context_processors.request',
                 'django.contrib.auth.context_processors.auth',
                 'django.contrib.messages.context_processors.messages',
+                'respa_admin.context_processors.export_global_vars',
```
jussih: Instead of a context processor (which affects all template rendering in the system, not just respa admin), this could alternatively be implemented as a Mixin class which is inherited in the Respa admin views.

Something like
```
class ExtraContextMixin():
    def get_context_data(self, **kwargs):
        context = super().get_context_data(**kwargs)
        context['INSTRUCTIONS_URL'] = settings.FOO
        return context
```

Then this could be used in Respa admin views like: `class SaveResourceView(ExtraContextMixin, CreateView):`

This way the data would only be injected to respa admin template contexts. I'm not saying this has to be changed, but if the global context processor is used, I'd use a more specific context variable name than `INSTRUCTIONS_URL`, maybe prefix like `RESPA_ADMIN_INSTRUCTIONS_URL`.

respa_admin/templates/respa_admin/_nav.html
```diff
@@ -9,6 +9,14 @@
         <!-- Collect the nav links, forms, and other content for toggling -->
         <div class="collapse navbar-collapse">
             <ul class="nav navbar-nav navbar-right">
+                {% if INSTRUCTIONS_URL %}
+                <li>
+                    <a href="{{ INSTRUCTIONS_URL }}">
```
jussih: Add target="_blank" here so the link opens in a new tab. Otherwise the user can lose unsaved data in the form if the link opens in the current tab.

PR #514, opened 2019-03-18
Document missing Reservation API endpoint filters

### LINKED ISSUE #513: Document start & end parameters for reservation endpoint
`start` and `end` parameters seem to work for reservation endpoint since
https://api.hel.fi/respa/v1/reservation/?resource=avmsgmrlb2ma&start=2019-01-01&end=2019-03-07
and
https://api.hel.fi/respa/v1/reservation/?resource=avmsgmrlb2ma&start=2018-01-01&end=2019-03-07
produce different results

Update swagger to reflect this. 

Also check what other parameters might be missing from the documentation. Some parameters might work with several endpoints but are only documented for one endpoint. So grab relevant parameters from one endpoint and see if they produce results with another endpoint.

### LINKED ISSUE #522: Reservation endpoint filter throws server error 500
Reservation endpoint filter `is_favorite_resource` contains a bug, which causes it to throw an exception when trying to return an empty queryset for unauthenticated user.

otiala:
Closes #513 
Closes #522 

Also fixes a minor bug described in the #522

### DIFF
--- a/resources/api/reservation.py
+++ b/resources/api/reservation.py
@@ -441,7 +441,7 @@ class ReservationFilterSet(django_filters.rest_framework.FilterSet):
         user = self.request.user
 
         if not user.is_authenticated:
-            return queryset.none if value else queryset
+            return queryset.none() if value else queryset
 
         filtering = {'resource__favorited_by': user}
         return queryset.filter(**filtering) if value else queryset.exclude(**filtering)
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -621,6 +621,14 @@ paths:
           in: query
           type: boolean
           description: Return only reservations that are related to resource that either is or is not favorited by user, based on the boolean value.
+        - name: user
+          in: query
+          type: string
+          description: Return only reservations that are related to given user
+        - name: is_own
+          in: query
+          type: boolean
+          description: Return only own reservations
       responses:
         '200':
           description: Successful response
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -624,7 +624,7 @@ paths:
         - name: user
           in: query
           type: string
-          description: Return only reservations that are related to given user
+          description: Return only reservations that are related to given user. Expects to recieve user's UUID.
         - name: is_own
           in: query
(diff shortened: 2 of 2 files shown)


swagger.yaml
```diff
@@ -605,6 +605,22 @@ paths:
...
+          type: string
+          description: The ending time of the reservation in ISO 8601 format.
+        - name: has_catering_order
+          in: query
+          type: boolean
+          description: Return only reservations which either have or do not have related catering orders based on the boolean value.
+        - name: is_favorite_resource
+          in: query
+          type: boolean
+          description: Return only reservations that are related to resource that either is or is not favorited by user, based on the boolean value.
```
jussih: I think there's still a couple filters missing. At least `user` and `is_own` from `UserFilterBackend`.

resources/api/reservation.py
```diff
@@ -441,7 +441,7 @@ def filter_is_favorite_resource(self, queryset, name, value):
         user = self.request.user
 
         if not user.is_authenticated:
-            return queryset.none if value else queryset
+            return queryset.none() if value else queryset
```
jussih: It's a bit surprising to find this change in a PR which judging by the title contains only documentation updates.

If this is a bug found while exploring parameters for the documentation and fixed here, maybe create an issue and link it in the PR and/or update the PR title or description to reflect there's an actual code fix included.

jussih:
The coverage requirement can be ignored here since the changes are to documentation.

resources/api/reservation.py
otiala: I created and issue and refered to it in the description.

swagger.yaml
```diff
@@ -621,6 +621,14 @@ paths:
           in: query
           type: boolean
           description: Return only reservations that are related to resource that either is or is not favorited by user, based on the boolean value.
+        - name: user
+          in: query
+          type: string
+          description: Return only reservations that are related to given user
```
jussih: This could specify better what the filter expects. It needs the UUID of the user to work.

PR #518, opened 2019-03-20
Order resources by translated type name instead of ID

### LINKED ISSUE #517: Order resources by translated name instead of ID
Order resources by translated name instead of ID

otiala:
Closes #517 

Order resources by translated type name instead of ID

### DIFF
--- a/resources/api/resource.py
+++ b/resources/api/resource.py
@@ -336,7 +336,9 @@ class ResourceFilterSet(django_filters.FilterSet):
             ('unit__name_fi', 'unit_name_fi'),
             ('unit__name_en', 'unit_name_en'),
             ('unit__name_sv', 'unit_name_sv'),
-            ('type', 'type'),
+            ('type__name_fi', 'type_fi'),
+            ('type__name_en', 'type_en'),
+            ('type__name_sv', 'type_sv'),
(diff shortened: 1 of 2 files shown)


resources/api/resource.py
```diff
@@ -336,7 +336,9 @@ def __init__(self, *args, **kwargs):
             ('unit__name_fi', 'unit_name_fi'),
             ('unit__name_en', 'unit_name_en'),
             ('unit__name_sv', 'unit_name_sv'),
-            ('type', 'type'),
```
tuomas777: You should not modify or remove any already existing fields to maintain backwards compatibility.

otiala: @tuomas777 no implementations on the frontend have been made yet, it's a brand new feature.

tuomas777: Ok, so that has not been even deployed yet? In that case it is probably ok :)

petterill: Changes like these should be discussed before they are made. In this case the change can be made since it is only deployed on the test server. If this feature had made it into production, we could not change it anymore. Also, Respa exists independent of Varaamo and has other clients, so whether or not it is implemented in Varaamo does not matter.

PR #529, opened 2019-04-04
Improve functions menu

### LINKED ISSUE #489: Improve Functions-menu
The menu at the bottom of the resource page (Toiminnot)

What it looks like now
toiminnot-respa-admin-nyt

What it could look like (from https://projects.invisionapp.com/share/NPIWNVYGBRU#/screens/297741145)

toiminnot-respa-admin

Specs:
- Add a "Tarkastele muutoksia / View changes" button. This link should open in a new window and take the user to the resource's Varaamo page. The link URL should also be configurable, so that it allows for Varaamo URL in other cities (or some other UI which is not Varaamo). Also, take into account that Respa Admin is used by Helsinki and Espoo employees (the same instance). Maybe just open the Helsinki version of Varaamo UI as default, it if's too difficult to open the Espoo URL when Espoo staff use Respa Admin.
- The button should only be visible for resources that are saved in the database. So not visible for new resources that have not yet been saved.
- Add the nice little coloured dots that are seen in the resources listing. So red, yellow and green dots in front of Julkaistu / Piilotettu / Varattavissa / Ei varattavissa
- Also remove the unnecessary colon after Toiminnot

Here https://projects.invisionapp.com/share/NPIWNVYGBRU#/screens/297741145 we have a similar button but it is for preview (which is not really possible).

otiala:
Closes #489 

- Added colorcoded ball icons to functions menu dropdown.
- Added a preview button to functions menu. Url to be redirected to is formed from `VARAAMO_RESOURCES_URL` of `.env` file + resource ID. `VARAAMO_RESOURCES_URL` is assumed to be something like `https://varaamourl.fi/resources/`


jussih:
I feel the `View changes` button could be bad UX. At least for me as user the instinct would be to make some changes in the form, then click view changes BEFORE pressing save.

Perhaps change the button label to `View resource`? I think it would be less misleading that way.

respa/settings.py
```diff
@@ -34,6 +34,7 @@
     MAIL_MAILGUN_DOMAIN=(str, ''),
     MAIL_MAILGUN_API=(str, ''),
     RESPA_IMAGE_BASE_URL=(str, ''),
+    VARAAMO_RESOURCES_URL=(str, ''),
```
jussih: Change this to a more generic name as there can be other clients than Varaamo.

Perhaps `RESPA_ADMIN_VIEW_RESOURCE_URL` ?

petterill:
> I feel the `View changes` button could be bad UX. At least for me as user the instinct would be to make some changes in the form, then click view changes BEFORE pressing save.
> 
> Perhaps change the button label to `View resource`? I think it would be less misleading that way.

Agreed. Change as Jussi suggested.

respa_admin/templates/respa_admin/resources/form/_toolbar.html
```diff
@@ -2,15 +2,22 @@
...
         <div class='select-dropdown btn btn-primary'>
-            {{ form.reservable }}
+            <div id='reservable-dropdown-icon'></div> {{ form.reservable }}
             <span class="glyphicon glyphicon-chevron-down" aria-hidden="true"></span>
         </div>
         <button class='btn btn-primary toolbar-save' type="submit">{% trans "Save" %}</button>
+        {% if resource and VARAAMO_RESOURCE_PREVIEW_URL %}
+            <a href="{{ VARAAMO_RESOURCE_PREVIEW_URL }}" target="_blank" rel="noopener noreferrer">
+                <button class='btn toolbar-preview' type='button'>
+                {% trans "View changes" %}
```
jussih: Change the label to "View resource".

PR #532, opened 2019-04-08
Update dev/Dockerfile to match the one used in prod

JsseL:
The Dockerfile used in production uses already Python 3.6 and
has hence slightly differently named packages for libgdal and postgres
client.

These changes makes the Docker dev setup to be up to date with prod.

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -1,10 +1,10 @@
-FROM python:3.5
+FROM python:3.6
 
 WORKDIR /usr/src/app
 
 ENV APP_NAME respa
 
-RUN apt-get update && apt-get install -y libgdal1h postgresql-client-9.4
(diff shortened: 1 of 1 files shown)


jussiarpalahti:
Where is this dev specific dockerfile used?

JsseL:
@jussiarpalahti It's used in `docker-compose.yml` on [this line](https://github.com/City-of-Helsinki/respa/blob/master/docker-compose.yml#L17).

jussiarpalahti:
@JsseL Sorry, I meant *why* dev needs separate dockerfile? Couldn't we use one dockerfile for *everything*?

JsseL:
@jussiarpalahti Most likely we could. No objections on that. It **might** even make things simpler to maintain. This PR would merely fix the existing dev setup. I could look into refactoring the setup next time I happen to play with the Dockerfiles.

PR #539, opened 2019-04-25
Update readme

petterill:
Formatting, issue tracking, documentation, table of contents, settings. 

Django settings is mostly copypasta from the excellent work by @vikoivun at https://github.com/City-of-Helsinki/linkedevents/blob/master/config_dev.toml.example with Respa-specific settings thrown in. I'm not 100% sure if all the settings are correct, so appreciated if someone can review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,11 +108,36 @@ python manage.py resources_import --all kirjastot
 
 
 ### Settings
+
+- `DEBUG`: Whether to run Django in debug mode. [Django setting](https://docs.djangoproject.com/en/2.2/ref/settings/#debug).
+- `SECRET_KEY`: Secret used for various functions within Django. This setting is mandatory for Django. [Django setting](https://docs.djangoproject.com/en/2.2/ref/settings/#secret-key).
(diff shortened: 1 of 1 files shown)


JsseL:
Yay! 👍 Good stuff.

README.md
```diff
@@ -86,11 +108,36 @@ python manage.py resources_import --all kirjastot
 
 
 ### Settings
-- `RESPA_IMAGE_BASE_URL`: Base URL used when building image URLs in email notifications. Example value: `'https://api.hel.fi'`.
```
vikoivun: Add where the settings are set:
Settings are done either by setting environment variables named after the setting or adding them to `.env`-file in the project root. .env-file syntax is similar to TOML-files (INI-files), ie. KEY=VALUE pairs. Project root is the directory where this README is.

vikoivun: You can also set settings in local_settings.py, which allows you to set any variables whatsoever. However, some of the settings documented here are named differently in settings.py, especially authentication variables.

petterill: Very good, sir!

PR #542, opened 2019-04-29
Fix language buttons changing size during onclick event

otiala:
Added a white border to language buttons that are not selected, to avoid buttons changing size during onclick event.

### DIFF
--- a/respa_admin/static_src/styles/form/navbar-form.scss
+++ b/respa_admin/static_src/styles/form/navbar-form.scss
@@ -36,10 +36,6 @@
         opacity: $static-opacity;
         align-self: center;
       }
-
-      .language-item-selected {
-        border: solid 1px black !important;
-      }
     
       .language-item {
         font-weight: bold;
@@ -47,7 +43,7 @@
         border-radius: 20px;
         line-height: $badge-line-height;
         background-color: #fff;
-        border: solid 1px #fff;
+        border: solid 1px transparent;
                 
         span.badge {
           background-color: #ff0000;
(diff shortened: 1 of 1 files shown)


respa_admin/static_src/styles/form/navbar-form.scss
```diff
@@ -38,7 +38,7 @@
       }
 
       .language-item-selected {
-        border: solid 1px black;
+        border: solid 1px black !important;
```
jussih: `!important` in CSS is usually bad practice.
It can be removed by changing this selector to `.language-item.language-item-selected` so it becomes more specific and the rule has higher priority.

Also change `border: ...` to `border-color: black`. That way other border settings are not overwritten here as we only need to change the color.

respa_admin/static_src/styles/form/navbar-form.scss
```diff
@@ -47,6 +47,7 @@
         border-radius: 20px;
         line-height: $badge-line-height;
         background-color: #fff;
+        border: solid 1px #fff;
```
jussih: Use `transparent` instead of `#fff`.

respa_admin/static_src/styles/form/navbar-form.scss
otiala: Good point. Changed the priority by changing the order of the rules.

respa_admin/static_src/styles/form/navbar-form.scss
otiala: Done!

PR #543, opened 2019-04-29
Implement Product version history

tuomas777:
Old versions of products need to be saved so that Orders that refer to those
won't change when the products change. This is achieved by implementing same
table versioning for products, ie. current and historical versions will live in
the same model / database table.

Products have product_id (int) field, which is the same for all the versions
of the same product. In addition, there is archived_at field, which contains
either sentinel value ARCHIVED_AT_NONE for product versions that are currently
in use, or a datetime when the version has been archived from use by either
soft deletion or being surpassed by a new version.


tuomas777:
Fixed the reported typos.

PR #547, opened 2019-05-08
Resource accessibility data importing and API

jussih:
Accessibility data can be imported from KAPA accessibility database via `python manage.py accessibility_import`.

Settings values used:
* `RESPA_ACCESSIBILITY_API_BASE_URL = "https://asiointi.hel.fi/kapaesteettomyys/"` in testing environments this should be `https://asiointi.hel.fi/kapaesteettomyys_testi`

API changes:
* The Resource endpoint can now optionally expose accessibility information via a query parameter `?include=accessibility_summaries`. The `include` parameter is implemented as a generic Serializer mixin class which can be used in other serializers where optional field includes are needed.
* Resource listing can be ordered by accessibility viewpoints. Query with `?order_by=-accessibility&accessibility_viewpoint=11` - the accessibility viewpoint id must be defined for ordering. The viewpoint IDs are the same as in the Accessibility API: https://asiointi.hel.fi/kapaesteettomyys/api/v1/accessibility/viewpoints
* Unit accessibility data can be queried by also requesting unit detail data by appending `include=unit_detail` to the query params
* Example with all the new stuff (order by a specific viewpoint, include both unit and resource accessibility summaries): `http://localhost:8000/v1/resource/?include=accessibility_summaries&include=unit_detail&order_by=-accessibility&accessibility_viewpoint=11`

Note:
* Requesting accessibility has a performance impact (the resource query is slower). Some optimization could be done to improve that. Normal queries are not affected.
* Ordering by multiple viewpoints is still missing and needs to be added later.


petterill:
@jussih is this WIP? Since importing unit accessibility data will be added too.

jussih:
@petterill Still WIP - ordering based on multiple viewpoints needs to be added based on what we decided. Updated the title.

Also "dummy" accessibility data with all values as unknown will be added for resources which have no accessibility data available.

PR #548, opened 2019-05-08
Change Order price API endpoint to work without a Reservation

tuomas777:
Instead of a reservation ID, the endpoint now requires a time range for the
price calculations, which is given by "begin" and "end" datetimes.

The change required some price helper refactoring, which hopefully also made
those somewhat cleaner and more usable.

Refs RESPA-66

### DIFF
--- a/payments/models.py
+++ b/payments/models.py
@@ -1,7 +1,6 @@
 import uuid
 from datetime import datetime, timedelta
-from decimal import ROUND_HALF_UP, Decimal
-from typing import Union
+from decimal import Decimal
 
 from django.core.validators import MaxValueValidator
 from django.db import models, transaction, IntegrityError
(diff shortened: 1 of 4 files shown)


payments/models.py
```diff
@@ -116,29 +115,34 @@ def _save_new(self, *args, **kwargs):
         else:
             raise Exception('Cannot obtain a new product ID for product {}.'.format(self))
 
-    def get_pretax_price_for_reservation(self, reservation: Reservation, rounded: bool = True,
-                                         as_sub_units: bool = False) -> Price:
+    def get_pretax_price_for_time_range(self, begin: datetime, end: datetime, rounded: bool = True) -> Decimal:
+        assert begin <= end
```
maosalo: Does it make sense to allow equality here?

tuomas777: My initial point was to disallow times that cause incorrect results, and when the times are the same, it still "works" by giving result 0. Then again, you are correct that using it with those equal values does not really make sense, so probably just better to disallow that.

tuomas777: Fixed.

PR #551, opened 2019-05-10
Add OpenAPI 3.0 specification

petterill:
- Converted the Swagger 2.0 spec to OpenAPI 3.0.1 spec with Swagger Editor (the editor does not support 3.0.2 yet)
- Rewrote descriptions and changed the API version to v1
- added example values for `resource` endpoint

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md

The Swagger 2.0 yaml file should probably be kept alongside this file (at least both versions are available in the [Servicemap API repo](https://github.com/City-of-Helsinki/smbackend) as well). But maintaining both files is maybe to much work, so at some point later on we could ditch the swagger file or at least mark it as deprecated.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1,0 +1,1253 @@
+openapi: 3.0.1
+info:
+  title: Respa
+  description: The Respa API provides categorized data on resources available for reservation within a city or metropolitan area and enables the reservation of these resources. The API provides data in the JSON format, in a RESTful fashion.
+  version: v1
+servers:
+- url: //api.hel.fi/respa/v1
+tags:
+- name: resource
+  description: Look for available resources
+- name: reservation
+  description: Make or change your reservations
+- name: unit
+  description: Places where resources are located
+- name: equipment
+  description: Equipment available for use in the resources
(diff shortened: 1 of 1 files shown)


otiala:
Validator seems to have a problem with line 198: `Can not read a block mapping entry; a multiline key may not be an implicit key`

openapi.yaml
```diff
@@ -0,0 +1,1252 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/purpose'
+  /type/:
+    get:
+      tags:
+      - filter
+      description: The type endpoint returns the possible resource types registered in the system. The optional parameter **page** 
+      allows specifying page number and **page_size** allows specifying more than the default 20 units per page.
+      parameters:
```
otiala: Validator seems to have a problem with line 198: `Can not read a block mapping entry; a multiline key may not be an implicit key`

petterill: Fixed by adding spaces. Validator seems happy now.

PR #552, opened 2019-05-13
Add missing information to notification context

otiala:
Went through the notification context variables.

Not working:
- `host_name`: It seems the client does not send this variable to backend. Needs to be fixed in Varaamo.

Added: 
- `event_description`
- `reserver_email_address`
- `reserver_phone_number`
- `authenticated_user_email`

### DIFF
--- a/resources/models/reservation.py
+++ b/resources/models/reservation.py
@@ -346,7 +346,9 @@ class Reservation(ModifiableModel):
                 'reserver_name': reserver_name,
                 'event_subject': self.event_subject,
                 'event_description': self.event_description,
-                'reserver_email_address': self.reserver_email_address, 
+                'reserver_email_address': self.reserver_email_address,
+                'reserver_phone_number': self.reserver_phone_number,
+                'authenticated_user_email': user.email
             }
             if self.resource.unit:
                 context['unit'] = self.resource.unit.name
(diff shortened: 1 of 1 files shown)


resources/models/reservation.py
```diff
@@ -345,6 +345,10 @@ def get_notification_context(self, language_code, user=None, notification_type=N
                 'host_name': self.host_name,
                 'reserver_name': reserver_name,
                 'event_subject': self.event_subject,
+                'event_description': self.event_description,
+                'reserver_email_address': self.reserver_email_address,
```
jussih: Can this be handled like `reserver_name`? If there is no `reserver_email_address` extra field set, use `user.email` if that is available.

resources/models/reservation.py
```diff
@@ -345,6 +345,10 @@ def get_notification_context(self, language_code, user=None, notification_type=N
...
+                'reserver_phone_number': self.reserver_phone_number,
+                'authenticated_user_email': user.email
```
jussih: If `reserver_email_address` is changed as above, this can be removed.

resources/models/reservation.py
petterill: Yes, implement it as @jussih suggested

PR #558, opened 2019-05-17
Remove webkit gradient from button and add condition to button rendering

otiala:
Remove webkit gradient from button and add condition to button rendering


petterill:
Yes, implement as Jussi suggested.

PR #563, opened 2019-05-20
RESPA-91: Add unit.id as locationId to accessibility data input link.

jussih:
Accessibility API input can now accept building id and it allows copying accessibility data between meeting rooms in the same location.

Unit.id is used as the locationId.

Discussion point:
Should we something other than Unit.id? If other apps than Respa use the Accessibility API it would be useful if the locationIds would match. Units can have many UnitIdentifiers which can contain IDs in other systems such as TPREK or HELMET. How would we then choose which identifier to send to Accessibility API?

### DIFF
--- a/respa_admin/accessibility_api.py
+++ b/respa_admin/accessibility_api.py
@@ -6,22 +6,27 @@ from hashlib import sha256
 from urllib.parse import quote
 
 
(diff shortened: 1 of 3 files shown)


jussih:
We can strip the `tprek:` prefix from the unit.id to use just the tprek id as the locationId so other systems can match the data that has been entered through respa.

PR #567, opened 2019-05-21
RESPA-96: Extend reservation endpoint resource filter to accept multiple values

tuomas777:


swagger.yaml
```diff
@@ -557,7 +557,7 @@ paths:
           type: integer
         - name: resource
           in: query
-          description: Resource id, for filtering reservations by resource
+          description: Resource id, for filtering reservations by resource. Accepts multiple comma-separated values.
```
jussih: Does this also work by providing multiples of the same query parameter?
As in `localhost:8000/v1/reservation?resource=abc&resource=123`

jussih:
The CI checks are blocked because the PR is coming from a fork. Looking at the code everything looks good though.

swagger.yaml
tuomas777: Pretty sure it does not, should it? I don't think other filters in Respa do that either (?)

tuomas777:
Travis build is now fixed.

PR #569, opened 2019-05-22
Try to skip docker image build, if password is not available

vikoivun:


vikoivun:
The assumption here is that DOCKER_USER is available to external PR:s, but DOCKER_PASSWORD is. Thus the script failed previously when it tried to login to docker hub.

PR #571, opened 2019-05-22
Use kirjastot.fi api v4 in library opening hour importer.

jussih:
Uses the new API v4 instead of v2 and v3 API used previously.

The new API can have several opening hours per day for a single library, for example:
- From 08:00 to 10:00 as self service
- From 10:00 to 12:00 with staff
- From 12:00 to 13:00 closed
- From 13:00 to 17:00 with staff

Currently the data model in Respa only allows for one opening time and one closing time per day. The workaround taken for now is that only staffed hours are looked at and any blocks in between are ignored, so the example above would end up as follows in Respa:
- Opens 10:00
- Closes 17:00

TODO before deployment: When syncing to a clean database, the Units are created with an identifier with namespace `kirkanta` - the old sync code only processed units in `helmet` and `kirjastot.fi` namespaces. Identifiers in production need to be checked.

TODO before deployment: Hours synced by the new version need to be compared to hours synced by the old version.

### DIFF
--- a/resources/importer/kirjastot.py
+++ b/resources/importer/kirjastot.py
@@ -1,20 +1,20 @@
+import calendar
 import datetime
-from collections import namedtuple
-import calendar, datetime
-
-import requests
-from django.core.exceptions import ObjectDoesNotExist, ValidationError
-from psycopg2.extras import DateRange
 import delorean
+import requests
+from collections import namedtuple
+from django.conf import settings
 from django.db import transaction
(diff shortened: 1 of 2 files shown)


jussih:
Leave the old importer script still in place so it can be used easily if problems should arise with V4 import.

jussih:
Failing check is from code coverage - it can be ignored in this case, no tests are done for importer code at the moment.

PR #587, opened 2019-06-12
Migrate from sentry/raven to sentry_sdk

vikoivun:
Raven has broken because of something to do with SSL. Switching to sentry_sdk unbreaks this.

Some importers had their own sentry captures. I think I translated their behaviour to sentry_sdk. See migrate importers commit.

### DIFF
--- a/respa/settings.py
+++ b/respa/settings.py
@@ -2,15 +2,22 @@
 Django settings for respa project.
 """
 
-# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 import os
+import subprocess
 import environ
 import sentry_sdk
-import subprocess
+from sentry_sdk.integrations.django import DjangoIntegration
 from django.utils.translation import ugettext_lazy as _
 from django.core.exceptions import ImproperlyConfigured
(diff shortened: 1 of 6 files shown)


respa/settings.py
```diff
@@ -5,11 +5,32 @@
...
 import environ
-import raven
+import sentry_sdk
+import subprocess
 from django.utils.translation import ugettext_lazy as _
 from django.core.exceptions import ImproperlyConfigured
+from sentry_sdk.integrations.django import DjangoIntegration
 
 
+def get_git_revision_hash():
```
jussih: Alternative without the git executable dependency would be from the old sdk: https://github.com/getsentry/raven-python/blob/d7d14f61b7fb425bcb15512f659626648c494f98/raven/versioning.py#L11

jussih: Another way without having to dig through the git repo would be to write the release name to a file during the execution of the release script. That would be faster than calling an external process during Django startup.

vikoivun: After a bit of back and forth, agreement has reach that `git describe --tags --long --always` is actually quite informative. I'll also pick up the revision file as backup.

PR #596, opened 2019-06-18
Implement extra questions for reservation

otiala:
Implement extra questions for reservation

- Add `reservation_extra_questions` attribute for both `Resource` and `Reservation` models
- Add data migration to create `ReservationMetadataField` for `reservation_extra_questions`
- Add Resource's `reservation_extra_questions` to Respa Admin forms
- Include `reservation_extra_questions` to existing test (`test_reservation_metadata_set`)


resources/models/reservation.py
```diff
@@ -118,6 +119,7 @@ class Reservation(ModifiableModel):
                                                               null=True)
     participants = models.TextField(verbose_name=_('Participants'), blank=True)
     host_name = models.CharField(verbose_name=_('Host name'), max_length=100, blank=True)
+    reservation_extra_questions = models.TextField(verbose_name=_('Reservation extra questions'), null=True, blank=True)
```
jussih: Is there a specific need for this to be `null=True`? Other TextFields here are only `blank=True`.

resources/models/resource.py
```diff
@@ -232,6 +232,7 @@ class Resource(ModifiableModel, AutoIdentifiedModel):
         verbose_name=_('External reservation URL'),
         help_text=_('A link to an external reservation system if this resource is managed elsewhere'),
         null=True, blank=True)
+    reservation_extra_questions = models.TextField(verbose_name=_('Reservation extra questions'), null=True, blank=True)
```
jussih: Same question here as above

resources/models/reservation.py
otiala: You're right, `null=True` on `TextField` is unnecessary and may cause confusion. I'll change it!

PR #608, opened 2019-07-01
payments.md: Add info about how to configure the payments app

maosalo:
Also a bit of general description and a couple links to the official API doc.

Refs RESPA-137

### DIFF
--- a/docs/payments.md
+++ b/docs/payments.md
@@ -2,6 +2,25 @@
 
 Payments app adds support for Respa resources to have paid reservations. Transactions are handled by a third party provider such as Bambora Payform.
 
+## Enabling and configuring Payments
+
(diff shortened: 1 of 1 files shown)


docs/payments.md
```diff
@@ -2,6 +2,25 @@
 
 Payments app adds support for Respa resources to have paid reservations. Transactions are handled by a third party provider such as Bambora Payform.
 
+## Enabling and configuring Payments
+
+There are a couple of required configuration keys that need to be set in order to enable the payments and the third party provider currently in use:
+
+- `RESPA_PAYMENTS_ENABLED`: Whether app is enabled or not. Boolean `True`/`False`. The default value is `False`.
```
tuomas777: I would rather put "Whether payments are enabled or not." or something similar, because technically the app is in use always.

maosalo: Alright, changed that one as suggested

PR #613, opened 2019-07-11
One request order creation

tuomas777:
Change the way paid reservations are created

Previously in order to create a reservation with an order, one needed
to first create the reservation, and then the order on a separate
request. From now on, orders are created using field "order" inside
reservation data when POSTing a new reservation. The "order" field
must contain all the same fields as before. The reservation endpoint
is only used to create the orders; when updating a reservation, the
"order" field is read-only.

In addition to the functional changes, carried out some refactoring
related to order / reservation serializers and importing of those,
the biggest change being that payments' api has been divided into
multiple modules.

Refs RESPA-139


payments/management/commands/update_expired_orders.py
```diff
@@ -0,0 +1,18 @@
...
+from django.core.management.base import BaseCommand
+from django.db.transaction import atomic
+
+from payments.models import Order
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Sets too old orders from state "waiting" to state "expired."'
```
maosalo: I suppose the dot at the end near "expired" should be outside of the quotation marks.

payments/tests/test_reservation_api.py
```diff
@@ -108,10 +154,112 @@ def test_reservation_that_has_order_cannot_be_modified_without_permission(user_a
...
+    order_data = build_order_data(product=product_with_quantity, quantity=quantity)
+    reservation_data['order'] = order_data
+
+    response = user_api_client.post(LIST_URL, reservation_data)
+
+    assert response.status_code == expected_status, response.data
+
+
+@pytest.mark.parametrize('has_product', (True, False))
+def test_order_mandatoriness(user_api_client, resource_in_unit, has_product):
```
maosalo: It seems "mandatoriness" is a real word although I don't think I've seen it used before. :)

payments/management/commands/update_expired_orders.py
tuomas777: 👍

payments/tests/test_reservation_api.py
tuomas777: Yeah, actually I needed to verify it was a real word as well :D

PR #614, opened 2019-07-12
Change provider init logic, rename order_create and move parts of the…

maosalo:
… URL logic into base

Provider configuration is now checked when app runs and provider instances are now tied to individual requests
Provider order_create renamed to initiate_payment that now only takes an Order object as param
The UI return URL is now created and parsed in base so that it isn't necessary to implement it in the subclass


payments/providers/base.py
```diff
@@ -59,7 +80,7 @@ def ui_redirect_success(cls, return_url: str, order: Order = None) -> HttpRespon
         This should be used after a successful payment instead of the
         standard Django redirect.
         """
-        return cls._redirect_to_ui(return_url, 'success', order)
+        return cls._redirect_to_ui(return_url, 'success', order) if return_url else HttpResponse(status=204)
```
tuomas777: The Respa 204 response could contain at least some text like "Payment success but error redirecting back to UI" or something. It won't be localized nor perhaps understandable to most of the end-users, but still better than a blank screen IMO.

payments/providers/base.py
```diff
@@ -68,7 +89,7 @@ def ui_redirect_failure(cls, return_url: str, order: Order = None) -> HttpRespon
         This should be used after a failed payment instead of the
         standard Django redirect.
         """
-        return cls._redirect_to_ui(return_url, 'failure', order)
+        return cls._redirect_to_ui(return_url, 'failure', order) if return_url else HttpResponseServerError()
```
tuomas777: Ditto, "Payment failure and cannot even redirect back to UI, damn oh damn."

payments/providers/bambora_payform.py
```diff
@@ -49,20 +45,20 @@ def get_config_template() -> dict:
...
         }
 
-    def order_create(self, request, ui_return_url, order) -> str:
+    def initiate_payment(self, order) -> str:
         """Initiate payment by constructing the payload with necessary items"""
 
-        respa_return_url = self.get_success_url(request)
-        query_params = urlencode({UI_RETURN_URL_PARAM_NAME: ui_return_url})
-        full_return_url = '{}?{}'.format(respa_return_url, query_params)
+        # request: HttpRequest, ui_return_url: str,
```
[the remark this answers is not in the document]
maosalo: Yeah, copied that as a reminder and forgot to remove. :)

payments/providers/base.py
maosalo: Yea you're right, although these cases where return url is missing should rarely happen it doesn't hurt if there's some message attached.

payments/providers/base.py
maosalo: Yup, same thing.

PR #617, opened 2019-08-04
Update dev env

igordavydsson:
- Ignore data/ and static/ folders in git
- Update webpack-cli version
- Update docker-compose setup

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,19 +3,19 @@ services:
   db:
     container_name: respa-db
     build: ./docker/postgres/
-    restart: always
+    restart: on-failure
     environment:
         POSTGRES_USER: respa
         POSTGRES_PASSWORD: respa
         POSTGRES_DB: respa
     volumes:
-      - db_data:/var/lib/postgresql
+      - db_data:/var/lib/postgresql/data
   api:
     container_name: respa-api
     build:
       context: ./
       dockerfile: ./docker/dev/Dockerfile
-    command: bash -c 'tail -f /dev/null'
+    command: python manage.py runserver 0:8000
     volumes:
       - .:/usr/src/app
     environment:
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,9 @@ __pycache__
 db.sqlite3
 *.mo
 *.sqlite
+data/
 media/
+static/
 .django_secret
 ./local_settings.py
 /venv*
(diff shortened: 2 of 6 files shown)


docker-compose.yml
```diff
@@ -3,19 +3,19 @@ services:
   db:
     container_name: respa-db
     build: ./docker/postgres/
-    restart: always
+    restart: on-failure
     environment:
```
pkalliok: I don't understand the ramifications for this.  Is it a developer help to not ditch the db contents upon docker-compose stop & docker-compose up?

docker-compose.yml
```diff
@@ -3,19 +3,19 @@ services:
...
         POSTGRES_USER: respa
         POSTGRES_PASSWORD: respa
         POSTGRES_DB: respa
     volumes:
-      - db_data:/var/lib/postgresql
+      - db_data:/var/lib/postgresql/data
   api:
```
pkalliok: This is probably OK, but I'm not sure if it affects anything.

docker-compose.yml
```diff
@@ -3,19 +3,19 @@ services:
...
     container_name: respa-api
     build:
       context: ./
       dockerfile: ./docker/dev/Dockerfile
-    command: bash -c 'tail -f /dev/null'
+    command: python manage.py runserver 0:8000
```
pkalliok: Does it really work to just run the server?  README.md says to run the server with a separate command https://github.com/City-of-Helsinki/respa/blob/master/README.md#installation-with-docker .  This is now not in sync with what is said in the README.

docker-compose.yml
pkalliok: What is not persisted anymore?

igordavydsson: without this change every time you stop and start the container, the db data is gone. the container creates a new and empty db.

docker-compose.yml
igordavydsson: this is not about the contents, this is about the container that does not shut down ever, unless you manually locally change the restart policy, which is super annoying

igordavydsson: also this restart policy starts the container automatically upon the launch of docker on the machine, which leads to the port being occupied if you then decide to start another project, where db container uses the same port

docker-compose.yml
igordavydsson: true, gets out of sync with the README.

igordavydsson: updated README 🙂

jussih:
@jussiarpalahti should take a look at the docker-compose changes and see if they affect the automatic deploys to the container-driven test environment.

PR #618, opened 2019-08-06
Extend reservation endpoint

igordavydsson:
- Add include parameter to GET /reservation/ endpoint

Refs VAR-140

### DIFF
--- a/resources/api/reservation.py
+++ b/resources/api/reservation.py
@@ -26,7 +26,8 @@ from resources.models.utils import generate_reservation_xlsx, get_object_or_none
 
 from ..auth import is_general_admin
 from .base import (
-    NullableDateTimeField, TranslatedModelSerializer, register_view, DRFFilterBooleanWidget
+    NullableDateTimeField, TranslatedModelSerializer, register_view, DRFFilterBooleanWidget,
+    ExtraDataMixin
 )
 
 User = get_user_model()
@@ -63,7 +64,7 @@ class UserSerializer(TranslatedModelSerializer):
         fields = ('id', 'display_name', 'email')
 
 
-class ReservationSerializer(TranslatedModelSerializer, munigeo_api.GeoModelSerializer):
+class ReservationSerializer(ExtraDataMixin, TranslatedModelSerializer, munigeo_api.GeoModelSerializer):
     begin = NullableDateTimeField()
     end = NullableDateTimeField()
     user = UserSerializer(required=False)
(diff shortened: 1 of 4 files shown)


resources/api/reservation.py
```diff
@@ -502,9 +512,12 @@ class ReservationViewSet(munigeo_api.GeoModelAPIView, viewsets.ModelViewSet, Res
         .prefetch_related('catering_orders').prefetch_related('resource__groups').order_by('begin', 'resource__unit__name', 'resource__name')
 
     serializer_class = ReservationSerializer
-    filter_backends = (DjangoFilterBackend, filters.OrderingFilter, UserFilterBackend, ReservationFilterBackend,
-                       NeedManualConfirmationFilterBackend, StateFilterBackend, CanApproveFilterBackend)
+    filter_backends = (DjangoFilterBackend, filters.OrderingFilter, filters.SearchFilter, UserFilterBackend,
+                       ReservationFilterBackend, NeedManualConfirmationFilterBackend, StateFilterBackend,
+                       CanApproveFilterBackend)
     filterset_class = ReservationFilterSet
+    search_fields = ('user__first_name', 'user__last_name', 'user__email',
```
jussih: What's the use case for the generic search filter? Currently the search filters for Reservation are in ReservationFilterSet (https://github.com/City-of-Helsinki/respa/blob/master/resources/api/reservation.py#L400) and partially overlapping what this new `search` filter is doing.

Can the use case be satisfied through ReservationFilterSet?

igordavydsson: ah, okay, I think I can extend `ReservationFilterSet` then, thanks! 🙂

resources/api/reservation.py
```diff
@@ -115,6 +116,15 @@ def __init__(self, *args, **kwargs):
             for field_name in required:
                 self.fields[field_name].required = True
 
+    def get_extra_fields(self, includes, context):
+        from .resource import ResourceSerializer
+
+        """ Define extra fields that can be included via query parameters. Method from ExtraDataMixin."""
+        extra_fields = {}
+        if 'resource' in includes:
```
jussih: `ResourceSerializer` has used `unit_detail` as the key to get the full unit data included, use the same notation here.

resources/api/reservation.py
igordavydsson: This will be introduced in a separate PR.

resources/api/reservation.py
igordavydsson: changed the keyword

igordavydsson:
Okay, thanks @tuomas777 and @jussih ! 🙌 I took a different approach to the problem now. `ReservationViewSet` has [`.select_related('resource', 'resource__unit')`](https://github.com/City-of-Helsinki/respa/blob/master/resources/api/reservation.py#L501), so performance should not be an issue for the specified fields.

PR #621, opened 2019-09-03
Reservation type

jussih:
Reservation has a `type` attribute.

Currently supported types are `normal` and `blocked`. Only unit managers or admins can create a reservation of type `blocked`.

Client systems can use the type information to display the reservation in a unique way.


resources/api/reservation.py
```diff
@@ -191,6 +191,10 @@ def validate(self, data):
             if not is_resource_manager:
                 raise ValidationError(dict(staff_event=_('Only allowed to be set by resource managers')))
 
+        if 'type' in data:
+            if data['type'] != Reservation.TYPE_NORMAL and not (is_resource_admin or is_resource_manager):
```
otiala: Should this check be explicitly for `Reservation.TYPE_BLOCKED` if there are possibly more types coming in the future?

jussih: I wanted this to blanket all choices besides normal, so that the developer who adds more types in the future won't accidentally let customers create types of reservation they shouldn't.

If some type of reservation is added that's usable by regular users, it will be blocked by this validator, and the developer is forced to think about validation and refactor this piece accordingly. :)

otiala: Good thinking, approved :)

PR #625, opened 2019-09-17
RESPA-147: Respa admin opening hours usability

jussih:
https://helsinkisolutionoffice.atlassian.net/browse/RESPA-147

Usability update to the opening hour editor in Respa Admin. More information is displayed in the period row without having to open the accordion. Copy to next button added to daily opening hour rows, which copies the opening and closing times to the next row. Styles updated and fixed.


jussih:
I moved CSS around a bit since styles were scattered all over the place.

PR #630, opened 2019-09-26
RESPA-158 | Respa admin unit listing

jussih:
https://helsinkisolutionoffice.atlassian.net/browse/RESPA-158

Respa Admin lists units the current user can manage. Units now have 2 new attributes, `data_source` and `data_source_hours`, set to the name of the source where data is imported for the unit itself or the opening hours. Unit list only allows editing if unit has no external data sources.

Actual edit view is not included, only listing.


respa_admin/templates/respa_admin/resources/_unit_list.html
```diff
@@ -0,0 +1,47 @@
...
+        </div>
+
+        <div class="unit-list container-fluid">
+            <ul class="list-group">
+                <li class="list-group-item sort-bar">
+                    <div class="pull-left empty-holder col-1"></div>
+                    <a class="col-2 btn order-by-link" href="?order_by={% if order_by == "name" %}-name{% else %}name{% endif %}">
+                        {% trans "Name" %} <span class="glyphicon glyphicon-sort" aria-hidden="true"></span>
+                    </a>
+                    <span class="col-3 order-by-link">
```
otiala: Should users be able to order results by `editable`-status?

jussih: That's a good question. I didn't figure that would be a useful usecase. Maybe @petterill can weigh in on this one?

petterill: Probably users are interested in editing a specific unit and not so much interested in all units that can be edited. So just having them in alphabetical order is most useful. No need to sort by editable status at this time.

PR #631, opened 2019-10-01
RESPA-171 | Support for first and last name search

otiala:
Changed `filter_reserver_info_search` method to also support both first name **and** last name. Changed method works as follows:

If space character has been found in query parameter (t.ex. oskari tiala), query parameter will be split by space, into `oskari` and `tiala`, and reservations are filtered with `first_name='oskari', last_name='tiala' OR first_name='tiala', last_name='oskari'`, as we really can’t know in which order the names are provided. The search is case insensitive.

### DIFF
--- a/resources/api/reservation.py
+++ b/resources/api/reservation.py
@@ -479,8 +479,14 @@ class ReservationFilterSet(django_filters.rest_framework.FilterSet):
         # restrict results to reservations the user has right to see
         queryset = queryset.extra_fields_visible(user)
 
+        fields = ('user__first_name', 'user__last_name', 'user__email',
+                  'reserver_name', 'reserver_email_address', 'reserver_phone_number')
+        conditions = []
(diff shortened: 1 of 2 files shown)


jussih:
Change PR to point to develop

resources/api/reservation.py
```diff
@@ -479,6 +479,18 @@ def filter_reserver_info_search(self, queryset, name, value):
         # restrict results to reservations the user has right to see
         queryset = queryset.extra_fields_visible(user)
 
+        # assume that first_name and last_name were provided if empty space was found
+        if ' ' in value:
```
jussih: These should be integrated to the `conditions` array below. As it stands having a search term with a space skips all the other fields, so a search `+358 50 123456` would not produce any results.

otiala: Done!

PR #643, opened 2019-10-22
RESPA-185 | Add permissions for unit manager

otiala:
Closes https://helsinkisolutionoffice.atlassian.net/browse/RESPA-185

Add the following permission for unit managers (with tests):

- can_make_reservations
- can_modify_reservations
- can_ignore_opening_hours
- can_view_reservation_access_code
- can_view_reservation_extra_fields
- can_access_reservation_comments
- can_view_reservation_catering_orders

### DIFF
--- a/resources/tests/test_reservation_api.py
+++ b/resources/tests/test_reservation_api.py
@@ -2174,6 +2174,5 @@ def test_manager_can_view_reservation_catering_orders(api_client, reservation, u
     api_client.force_authenticate(user=unit_manager_user)
 
     response = api_client.get(detail_url)
-    # import pdb; pdb.set_trace()
     assert response.status_code == 200
     assert response.data['has_catering_order'] == True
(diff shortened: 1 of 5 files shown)


resources/tests/test_reservation_api.py
```diff
@@ -2065,3 +2065,115 @@ def test_reservation_cannot_add_bogus_type(resource_in_unit, reservation_data, a
...
+    CateringOrder.objects.create(
+        reservation=reservation,
+        provider=provider
+    )
+
+    detail_url = reverse('reservation-detail', kwargs={'pk': reservation.pk})
+    api_client.force_authenticate(user=unit_manager_user)
+
+    response = api_client.get(detail_url)
+    # import pdb; pdb.set_trace()
```
tuomas777: A leftover.

otiala:
> Also if you didn't already, you might want to check how many database queries this new implementation causes especially when requesting a list of resources. I might be wrong, but it also might be a problem :)

`print(len(connection.queries))` in `get_queryset`of `ResourceListViewSet`returns the same amount of queries as with the old implementation, but I'm not sure that's too reliable. Something like django-debug-toolbar could be useful as a dev dependency for cases like this.

tommimanbytes:
Commits of this pull request has already been merged with a previous PR. Closing.

PR #646, opened 2019-11-04
RESPA-204 | Anonymize user data

otiala:
https://helsinkisolutionoffice.atlassian.net/browse/RESPA-204

Add custom admin action to enable anonymizing user objects and related reservation objects. Anonymized fields are listed in Jira ticket.

### DIFF
--- a/resources/tests/test_admin_actions.py
+++ b/resources/tests/test_admin_actions.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+import pytest
+from django.contrib.auth import get_user_model
+
+from resources.models import Reservation
+from users.admin import pseudonymize_user_data
+
+
+@pytest.mark.django_db
+def test_pseudonymize_user_data(api_client, resource_in_unit, user):
+    """
+    Test pseudonymization of user data.
+    """
+    user.first_name = 'testi_ukkeli'
+    user.save()
+    Reservation.objects.create(
+        resource=resource_in_unit,
+        begin='2015-04-04T09:00:00+02:00',
+        end='2015-04-04T10:00:00+02:00',
+        user=user,
+        reserver_name='John Smith',
+        event_subject="John's welcome party",
+        state=Reservation.CONFIRMED
+    )
+    # pseudonymize_user_data expects a queryset instead of single object
+    test_user = get_user_model().objects.all().filter(first_name='testi_ukkeli')
+    pseudonymize_user_data(modeladmin=None, request=None, queryset=test_user)
+    assert get_user_model().objects.filter(first_name='testi_ukkeli').count() == 0
+    assert Reservation.objects.filter(resource=resource_in_unit)[0].event_description == 'Sensitive data of this reservation has been pseudonymized by a script.'
--- a/resources/tests/test_admin_actions.py
+++ b/resources/tests/test_admin_actions.py
@@ -19,7 +19,7 @@ def test_anonymize_user_data(api_client, resource_in_unit, user):
         end='2015-04-04T10:00:00+02:00',
         user=user,
         reserver_name='John Smith',
-        event_subject="John's welcome party",
+        event_subject='John\'s welcome party',
         state=Reservation.CONFIRMED
     )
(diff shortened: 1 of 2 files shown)


users/admin.py
```diff
@@ -18,6 +69,32 @@ def modify_field_data(field_data):
...
+def pseudonymize_user_data(modeladmin, request, queryset):
+    for user in queryset:
+        user.first_name = random.choice(first_names_list)
+        user.last_name = random.choice(last_names_list)
+        user.username = f'pseudonymized-{uuid.uuid4()}'
+        user.email = f'{user.first_name}.{user.last_name}@pseudonymized.net'.lower()
+        user.save()
+
+        user_reservations = Reservation.objects.filter(user=user)
+        if user_reservations:
```
tuomas777: This `if` seems unnecessary.

otiala: Changed

resources/tests/test_admin_actions.py
```diff
@@ -0,0 +1,29 @@
...
+        resource=resource_in_unit,
+        begin='2015-04-04T09:00:00+02:00',
+        end='2015-04-04T10:00:00+02:00',
+        user=user,
+        reserver_name='John Smith',
+        event_subject="John's welcome party",
+        state=Reservation.CONFIRMED
+    )
+    # anonymize_user_data expects a queryset instead of single object
+    test_user = get_user_model().objects.all().filter(first_name='testi_ukkeli')
```
tommimanbytes: Object manager method call `.all()` before using filter is unnecessary here.

users/admin.py
```diff
@@ -1,10 +1,61 @@
...
+    'Luna',
+    'Ronald',
+    'Bryanna',
+    'Augustus',
+    'Laurel',
+    'Miles',
+    'Patricia'
+    'Beckett'
+    'Elle'
+]
```
tommimanbytes: Missing commas here causing last element to be `'PatriciaBeckettElle'`.

resources/tests/test_admin_actions.py
```diff
@@ -0,0 +1,29 @@
...
+    anonymize_user_data(modeladmin=None, request=None, queryset=test_user)
+    assert get_user_model().objects.filter(first_name='testi_ukkeli').count() == 0
+    assert Reservation.objects.filter(resource=resource_in_unit)[0].event_description == 'Sensitive data of this reservation has been anonymized by a script.'
```
tommimanbytes: I think `.filter(resource=resource_in_unit)[0]` could be replaced with `.get(resource=resource_in_unit)`

resources/tests/test_admin_actions.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    user.first_name = 'testi_ukkeli'
+    user.save()
+    Reservation.objects.create(
+        resource=resource_in_unit,
+        begin='2015-04-04T09:00:00+02:00',
+        end='2015-04-04T10:00:00+02:00',
+        user=user,
+        reserver_name='John Smith',
+        event_subject="John's welcome party",
```
tommimanbytes: Double quotation mark `"` used instead of `'`. Should be consistent.

resources/tests/test_admin_actions.py
tommimanbytes: Line length here is also too long (155 chars) for pep8 spec (max 120 chars). Could do something like:

`anonymized_field_text = 'Sensitive data of this reservation has been anonymized by a script.'
assert Reservation.objects.get(resource=resource_in_unit).event_description == anonymized_field_text`

resources/tests/test_admin_actions.py
otiala: This is an exception due to single quote being used inside the string.

tommimanbytes: I think escaping the single quote would be more consistent approach. 

`'John\'s welcome party'`

otiala:
> The uuid field should perhaps be anonymized. It is a reference to the uuid field in Tunnistamo/the master authentication database.

Good point, changed code.

PR #648, opened 2019-11-12
RESPA-207 | Custom label for metadata set in respa-admin

otiala:
Added `custom_label` for generic `_select.html` component that enables overriding the label inherited from the model.


petterill:
@otiala I think we could lose the "Valitse" here. It would be more in line with the other headlines in the edit resource view.

otiala:
> @otiala I think we could lose the "Valitse" here. It would be more in line with the other headlines in the edit resource view.

Done!

respa_admin/templates/respa_admin/forms/_select.html
```diff
@@ -1,7 +1,11 @@
 <div class="form-group {% if not inline %}select-with-label{% endif %}">
     {% if not skip_label %}
         <label class="control-label input-label">
-            {{ field.label }}{% if field.field.required %}*{% endif %}
+            {% if custom_label %}
+                {{ custom_label }}{% if field.field.required %}*{% endif %}
+            {% else %}
+                {{ field.label }}{% if field.field.required %}*{% endif %}
+            {% endif %}
```
tommimanbytes: To prevent declearing `{% if field.field.required %}*{% endif %}` twice these lines could be refactored to 

```

{% if custom_label %} {{ custom_label }} {% else %} {{ field.label }} {% endif %}

{% if field.field.required %}*{% endif %}


```

PR #649, opened 2019-11-21
RESPA-177 | Enable Unit Admins and Unit Managers to bypass payment

otiala:
Enable Unit Admins or Unit Managers to bypass payment.

Created a new permission `can_bypass_payment`. The permission checking method which returns a boolean returns `True` for Admins and Managers (as in UA and UM).

`get_user_permissions()` method in `ResourceSerializer` now adds `can_bypass_payment` boolean to the `user_permissions` object, which will be used in Varaamo.


resources/tests/test_resource_api.py
```diff
@@ -84,19 +88,57 @@ def test_user_permissions_in_resource_endpoint(api_client, resource_in_unit, use
...
+                            can_bypass_payment=False)
 
     assign_perm('unit:can_ignore_opening_hours', group, resource_in_unit.unit)
     api_client.force_authenticate(user=user)
     _check_permissions_dict(api_client, resource_in_unit, is_admin=False,
-                            can_make_reservations=True, can_ignore_opening_hours=True)
+                            can_make_reservations=True, can_ignore_opening_hours=True,
+                            can_bypass_payment=False)
+
+    # user has explicit permission to bypass payment
```
tuomas777: This comment seems to be wrong?

otiala: What do you mean? Do you mean it should be `group has explicit permission`?

PR #650, opened 2019-11-28
Fix bug when syncing resources to the exchange

Jogha:
When creating new resource as a admin to the respa, the exchange syncing
does not work properly. System is syncing the resource to the exchange
before the resource is saved to the database and the sync function wants
to lock the resource in the database before doing it. This pattern
creates problem when the resource does not have any id, but at case of
the new resource we don't need to lock the database resource.

Fixing the problem with conditionally locking the database resource,
meaning that if the given instance does not have id, we skip the
database lock.


petterill:
@Jogha is this running in production for the Huonevaraus Respa instance? Have you tested this?

petterill:
@Jogha also, can you point this to develop instead of master

otiala (approved):

petterill:
OK this is already in production for Huvaja, so this can be merged.

PR #653, opened 2019-12-17
Disable login for inactive users

otiala:
For some reason helusers overrides DRF’s `authenticate_credentials` method without calling `super()`, thus bypassing the check for inactive users. More information in the Jira issue.

1. Create a new authentication class that inherits heluser's authentication and applies `is_active` check as it is in DRF by default.

2. Add tests

3. Slight refactoring of `JWTMixin` to make it reusable in other tests that require JWT auth.

### DIFF
--- a/resources/authentication.py
+++ b/resources/authentication.py
@@ -1,0 +1,18 @@
+from django.utils.translation import ugettext as _
+from rest_framework import exceptions
+from rest_framework_jwt.settings import api_settings
+
+from helusers.jwt import JWTAuthentication
+
+jwt_get_username_from_payload = api_settings.JWT_PAYLOAD_GET_USERNAME_HANDLER
+
+
+class RespaJWTAuthentication(JWTAuthentication):
+    def authenticate_credentials(self, payload):
+        user = super().authenticate_credentials(payload)
+
+        if user and not user.is_active:
+            msg = _('User account is disabled.')
+            raise exceptions.AuthenticationFailed(msg)
+
+        return user
--- a/resources/authentication.py
+++ b/resources/authentication.py
@@ -1,10 +1,7 @@
 from django.utils.translation import ugettext as _
 from rest_framework import exceptions
-from rest_framework_jwt.settings import api_settings
 
 from helusers.jwt import JWTAuthentication
-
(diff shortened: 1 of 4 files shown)


resources/authentication.py
```diff
@@ -0,0 +1,18 @@
+from django.utils.translation import ugettext as _
+from rest_framework import exceptions
+from rest_framework_jwt.settings import api_settings
+
+from helusers.jwt import JWTAuthentication
+
+jwt_get_username_from_payload = api_settings.JWT_PAYLOAD_GET_USERNAME_HANDLER
```
tommimanbytes: Unused variable?

resources/authentication.py
```diff
@@ -0,0 +1,18 @@
...
+
+class RespaJWTAuthentication(JWTAuthentication):
+    def authenticate_credentials(self, payload):
+        user = super().authenticate_credentials(payload)
+
+        if user and not user.is_active:
+            msg = _('User account is disabled.')
+            raise exceptions.AuthenticationFailed(msg)
+
+        return user
```
tommimanbytes: No new line at end

tommimanbytes:
This feature is now implemented in new version (0.5.3) of django-helusers. We could still use the test here. Also I tried updating django-helusers to newest version but that caused this error:

>   File "/home/tommi/yritys_projektit/haltu/respa/respa/urls.py", line 12, in <module>
    admin.autodiscover()
AttributeError: module 'helusers.admin' has no attribute 'autodiscover'

So I think this branch should update django-helusers library, resolve errors caused by the update and implement the test.

otiala:
> This feature is now implemented in new version (0.5.3) of django-helusers. We could still use the test here. Also I tried updating django-helusers to newest version but that caused this error:
> 
> > File "/home/tommi/yritys_projektit/haltu/respa/respa/urls.py", line 12, in 
> > admin.autodiscover()
> > AttributeError: module 'helusers.admin' has no attribute 'autodiscover'
> 
> So I think this branch should update django-helusers library, resolve errors caused by the update and implement the test.

I'll close this branch and start a new one :) Thanks!

PR #654, opened 2019-12-20
RESPA-220 | Database query optimizations for resource and reservation API endpoints

tommimanbytes:
Various query optimizations. Most of them are about preventing additional db queries by not using querysets `.filter()` or `.exists()` methods in places where they could raise query count cumulatively  in iterations. Also prefetching user models unit permissions.

**This optimization optimizes list endpoint queries of reservation and resource from this:**
Selection_071

**To this:**
Selection_072


resources/auth.py
```diff
@@ -13,7 +13,36 @@ def is_general_admin(user):
 
 
 def is_any_admin(user):
-    return is_authenticated_user(user) and (
-        is_general_admin(user) or
-        user.unit_group_authorizations.admin_level().exists() or
-        user.unit_authorizations.admin_level().exists())
+    from .enums import UnitGroupAuthorizationLevel, UnitAuthorizationLevel
```
otiala: Is there a specific reason for the imports at function level?

tommimanbytes: Not really. I changed them to be at module level.

jussiarpalahti:
Apropos, where do you got the pictures with timings and query counts, @tommimanbytes ?
 Looks very useful.

tommimanbytes:
> Apropos, where do you got the pictures with timings and query counts, @tommimanbytes ?
> Looks very useful.

I used Django Silk library. It's a library that records all requests made to API and provides a web UI with detailed info about database queries made during the request.

[Django Silk Github page](https://github.com/jazzband/django-silk)

PR #655, opened 2020-01-09
RESPA-226 | Add manager role to user permissions

otiala:
**This PR is built on top of #654, which must be merged before this one!**

Add manager role to user permissions object which is exposed in resource list endpoint. Add necessary tests.


tommimanbytes:
Looks good!

PR #667, opened 2020-01-24
RESPA-195 | Add .env.example and update README.md

otiala:
Add .env.example and update README.md


tommimanbytes:
Maybe this should be more like this file in linked events: [Example config file](https://github.com/City-of-Helsinki/linkedevents/blob/master/config_dev.toml.example)

I think it would be helpful if every possible configuration variable in settings would be present here and described with comments like in the link above. This would help new developers understand different settings in Respa. I don't think it adds much value as is.

PR #672, opened 2020-01-30
Kulkunen SiPass driver changes

petterill:
Recent changes in the SiPass API broke Kulkunen. Added the `ManagerEmailAddress` attribute and changed the `Status` value to 62 to fix this as per instructions from Kuva SiPass expert.


kulkunen/drivers/sipass.py
```diff
@@ -536,6 +536,7 @@ def api_create_cardholder(self, data):
                     'PagerNumber': '',
                     'PagerServiceProviderId': '0',
                     'PhoneNumber': ''
```
tommimanbytes: Comma missing from end of this line

petterill: ah yes, fixed

PR #674, opened 2020-02-05
Fix admin bypassing time restrictions

otiala:
When calling model's `clean()` method through serializer, admin was undefined and therefore every user would be treated as non-admin.

### DIFF
--- a/resources/models/reservation.py
+++ b/resources/models/reservation.py
@@ -346,9 +346,10 @@ class Reservation(ModifiableModel):
         that it can be excluded when checking if the resource is available.
         """
 
-        user = self.user
-        if not user and 'user' in kwargs:
+        if 'user' in kwargs:
             user = kwargs['user']
+        else:
+            user = self.user
 
         user_is_admin = user and self.resource.is_admin(user)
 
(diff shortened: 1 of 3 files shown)


resources/models/reservation.py
```diff
@@ -346,7 +346,11 @@ def clean(self, **kwargs):
         that it can be excluded when checking if the resource is available.
         """
 
-        user_is_admin = self.resource.is_admin(self.user)
+        user = self.user
+        if not user and 'user' in kwargs:
```
tommimanbytes: So user in kwargs (request user) is used only when reservation has no user set? Isn't use case where admin can make reservation for another user and bypass min period valid?

otiala: Serializer is calling `clean()` method of the instance, which is pretty uncommon in DRF, but when doing that, we're not aware of the request nor the user in the clean method, so we're passing the user in kwargs. This way we're able to decide if the user is admin or not, if we're calling the model's clean method or accessing it through the serializer.

The reason it didn't work before is `self.user` would return `None` when accessing `clean()` method through serializer.

tommimanbytes: User in kwargs is the request user as it is taken from context['request'] in the serializer. Therefore you are checking if reservation objects related user is admin, not the user that is actually making this request. I think that you should first check if user in kwargs is present (request user) and check their role and if not then check the self.user (reservations related user). It is a question about whether to check if user editing / creating the reservation has permission to bypass things or if the user whose reservation it is has the permissions. I think the first option is more valid.

PR #687, opened 2020-02-26
RESPA-197 | Added payment terms field to Resource model and API.

tommimanbytes:
Also added the field to Respa Admin and tests. Payment terms is similar to generic terms and uses TermsOfUse model. 

There is now type field in TermsOfUse to filter only certain types of terms to lists in Respa Admin:

image

Type is also represented in the list view of Django admin:

image

This PR also closes following issue: [RESPA-212](https://helsinkisolutionoffice.atlassian.net/secure/RapidBoard.jspa?rapidView=27&projectKey=RESPA&modal=detail&selectedIssue=RESPA-212)


resources/models/resource.py
```diff
@@ -205,7 +214,9 @@ class Resource(ModifiableModel, AutoIdentifiedModel):
     reservation_info = models.TextField(verbose_name=_('Reservation info'), null=True, blank=True)
     responsible_contact_info = models.TextField(verbose_name=_('Responsible contact info'), blank=True)
     generic_terms = models.ForeignKey(TermsOfUse, verbose_name=_('Generic terms'), null=True, blank=True,
-                                      on_delete=models.SET_NULL)
+                                      on_delete=models.SET_NULL, related_name='resources_where_generic_terms')
+    payment_terms = models.ForeignKey(TermsOfUse, verbose_name=_('Payment terms'), null=True, blank=True,
+                                      on_delete=models.SET_NULL, related_name='resources_where_payment_terms')
     specific_terms = models.TextField(verbose_name=_('Specific terms'), blank=True)
```
otiala: Should `specific_terms` be refactored as a type of `TermsOfUse` class as well?

tommimanbytes: `generic_terms` and `payment_terms` uses predefined set of terms defined in Django Admin. These predefined terms can then be chosen to certain resource from a drop down menu in Respa Admin by admins and managers. Same terms are used for multiple resources.

`specific_terms` does not pick its value from any predefined set of terms because they are specific for a certain resource. Therefore defining them separately as TermsOfUse objects would in my opinion only add complexity and no positive effects that I can think of.

otiala: Thanks for clarification. Seems reasonable. Approved :)

PR #693, opened 2020-03-06
RESPA-254 | Disable adding new images for unsaved resources

otiala:
- Disable "Add image" button if resource is not saved
- Display disabled button according to Helsinki graphical guides


otiala:
Fixed in #700. I'm closing this one.

PR #697, opened 2020-03-12
RESPA-223 | Prevent overlapping reservations

otiala:
Prevent overlapping reservations on Unit-level, so that user won't be able to create two reservations of different resource in the same time-window.


resources/models/reservation.py
```diff
@@ -364,6 +364,21 @@ def clean(self, **kwargs):
             if day and not is_valid_time_slot(dt, self.resource.slot_size, day['opens']):
                 raise ValidationError(_("Begin and end time must match time slots"), code='invalid_time_slot')
 
+        # Check if Unit has disallow_overlapping_reservations value of True
+        if (
+            self.resource.unit.disallow_overlapping_reservations and not
+            self.resource.can_create_overlapping_reservations(user)
+        ):
+            reservations_for_same_unit = Reservation.objects.filter(user=user, resource__unit=self.resource.unit)
```
tommimanbytes: We should probably check if overlapping reservations are active (not cancelled). We probably want to allow a situation where users old reservation is cancelled and he then makes a new reservation for that time period.

otiala: Good catch. Changed!

PR #729, opened 2020-08-05
RESPA-277 | Add a new feature that allows cancelling reservations with predefined and custom reasons

tommimanbytes:
Add a new feature that allows cancelling reservations with predefined and custom reasons.

Reservation cancellation reason categories are first defined in Django Admin in all three languages. Cancellation reason categories all includes a predefined description text which is then added to cancellation / declination mails if reservation gets cancelled. Cancellation reasons are also shown to user in UI. The user which cancels the reservation can also type a custom cancellation message to reservation in addition to categorys predefined one.

Usage example

1. Define reservation cancellation reason categories in Django Admin
2. Cancel a reservation using PATCH HTTP-method with one of the defined categories and custom message: 
```json
{
  "state": "cancelled",
  "cancel_reason": {
    "description": "This reservation has been cancelled due to x and y",
    "category_id": 1
  }
}
```
3. Show cancel reason in UI to users and staff. Representation of cancel reason in reservation object in API:
```json
"cancel_reason": {
    "category": {
      "id": 1,
      "reservation_type": "confirmed",
      "name": {
        "fi": "grtre",
        "en": "gregr",
        "sv": "greg"
      },
      "description": {
        "fi": "regerg",
        "en": "egerg",
        "sv": "ergre"
      }
    },
    "description": "This reservation has been cancelled due to x and y",
  }
```


openapi.yaml
```diff
@@ -1496,3 +1561,53 @@ components:
...
+          description: "Type of reservation for which this cancel reason category is used. Choices: confirmed, requested, own."
+        name:
+          type: object
+          properties:
+            sv:
+              type: string
+              description: Swedish name for the cancel reason category
+            fi:
+              type: string
+              description: Swedish name for the cancel reason category
```
indigane, line 1598: Swedish in all fields.

resources/api/reservation.py
```diff
@@ -167,7 +193,7 @@ def validate(self, data):
         if not resource.can_make_reservations(request_user):
             raise PermissionDenied(_('You are not allowed to make reservations in this resource.'))
 
-        if data['end'] < timezone.now():
+        if data.get('end', False) and data['end'] < timezone.now():
```
indigane, line 196: `if 'end' in data` is more explicit for testing key existence than `data.get()`. Mostly matters for readability. I had to think for a bit why `.get` is used before I understood this line.

openapi.yaml
tommimanbytes: Fix'd

resources/api/reservation.py
tommimanbytes: Agreed, fix'd

PR #740, opened 2020-09-28
Add order_number field under order object in reservation API endpoint

tommimanbytes:
To get full order object with reservation, use _?include=order_detail_ GET-parameter


tommimanbytes:
Not required.

PR #756, opened 2021-03-03
Helhuv 34/import resources command

pouryadel:
Import resources via CSV file command added.
CSV file of dummy data to using the command added.

There might be data in the exchange resource column


resources/management/commands/resources_spaces_import.py
```diff
@@ -0,0 +1,79 @@
...
+        parser.add_argument(
+            "file_path", help="Path to the csv file",
+        )
+
+    @staticmethod
+    def get_authentication_type(authentication_type_fi):
+        if authentication_type_fi == 'ei mitään':
+            return 'none'
+        with translation.override("fi"):
+            name = [k for k in dict(Resource.AUTHENTICATION_TYPES).keys() if
```
nicobav, line 23: would make this a bit readable if the `dict(Resource.AUTHENTICATION_TYPES)` would be declared to its own variable: `authentication_types = dict(Resource.AUTHENTICATION_TYPES)`

resources/management/commands/resources_spaces_import.py
```diff
@@ -0,0 +1,79 @@
...
+    def handle(self, *args, **options):
+        file_path = options["file_path"]
+        if not os.path.exists(file_path):
+            raise CommandError("File {0} does not exist".format(file_path))
+
+        with open(file_path) as f:
+            csv_reader = csv.reader(f, delimiter=';')
+            print('Processing...')
+            for idx, row in enumerate(csv_reader):
+                if idx == 0 or not row[4]:
```
nicobav, line 36: Would be nice if the column indexes would be named in a dict (or enum/const) so that they would be accessible like `row[columns["title"]]` (dict)  or `row[COLUMNS.TITLE]` (enum/const)

resources/management/commands/resources_spaces_import.py
```diff
@@ -0,0 +1,79 @@
...
+    help = "Import resources via csv file. This is for importing resources for spaces"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "file_path", help="Path to the csv file",
+        )
+
+    @staticmethod
+    def get_authentication_type(authentication_type_fi):
+        if authentication_type_fi == 'ei mitään':
```
nicobav, line 20: consistence while using quotes or doulbe quotes => I suggest using double quotes `"ei mitään"`

resources/data/dummy_data_import_resources.csv
```diff
@@ -0,0 +1,31 @@
+﻿Julkinen;Toimipiste;Resurssityyppi;Käyttötarkoitukset;Nimi [fi];Kuvaus [fi];Tunnistautuminen;Henkilömäärä;Pinta-ala;Lyhin varausaika;Pisin varausaika;Varattavissa;Varausinfo [fi];Resurssiryhmät
```
nicobav, line 1: This file could be put unders tests and test_data or something like.

resources/management/commands/resources_spaces_import.py
pouryadel: Thanks. I'll change it

resources/management/commands/resources_spaces_import.py
pouryadel: I'll check it

aprenfors (approved):

aprenfors (commented):
Looks like there are some test failures, are those related to this?

pouryadel:
> Looks like there are some test failures, are those related to this?

I'm not sure, but I don't think that tests are related to this. Those are API tests that are not related to this MR

aprenfors (approved):

PR #819, opened 2023-04-27
Use postgres 12 on local dev db

nicobav:
Old postgres 10 seems not to work properly anymore


saaste (approved):
Awesome! 👏 

Same issue with SonarCloud as with the other PR.

PR #820, opened 2023-04-27
Use python 3.8 image

nicobav:


saaste (approved):
Great! 👏 

I think you need to manually review those SonarCloud issues. I tried to do it but I couldn't find an approve button. Perhaps because it is your PR and the issues were assigned to you? 🤔

PR #825, opened 2023-06-15
Use python:3.8-bullseye as base image

nicobav:
And ugprade node to 14.

The bullseye supports olders nodes unlike the base image 3.8 which removed bullseye in favor of bookworm.


saaste (approved):
🚀

PR #828, opened 2023-10-23
Update deps

matti-lamppu:
Updates dependencies where [dependabot is complaining about insecure versions](https://github.com/City-of-Helsinki/respa/security/dependabot).


matti-lamppu:
The only one which I couldn't fully fix is [Prototype Pollution in Ajv ](https://github.com/City-of-Helsinki/respa/security/dependabot/36), since [extract-text-webpack-plugin](https://www.npmjs.com/package/extract-text-webpack-plugin) is not maintained and requires an older version.

matti-lamppu:
If you look at the CI pipelines, it says that there are failing tests, but I can't replicate this locally, and all tests pass there, so I think it's OK to merge. Seem like this has [been a problem in CI for some time](https://dev.azure.com/City-of-Helsinki/respa/_build/results?buildId=103209&view=results).

vergama:
> The only one which I couldn't fully fix is [Prototype Pollution in Ajv ](https://github.com/City-of-Helsinki/respa/security/dependabot/36), since [extract-text-webpack-plugin](https://www.npmjs.com/package/extract-text-webpack-plugin) is not maintained and requires an older version.

I think we can live with that.

vergama:
> If you look at the CI pipelines, it says that there are failing tests, but I can't replicate this locally, and all tests pass there, so I think it's OK to merge. Seem like this has [been a problem in CI for some time](https://dev.azure.com/City-of-Helsinki/respa/_build/results?buildId=103209&view=results).

Looks like that; maybe we'll accept that too while Respa is still in use, and test after deploy.

vergama (approved):

PR #832, opened 2023-11-08
Use .get with default to get viewpointOrderText in accessibility import

matti-lamppu:
Try to fix reported cron issues. The current fix simply applies a default if the value is not found, but this might not be the correct thing to do.


vergama (approved):
Not clear if this was the underlying issue since the accessibility URL now returns only elements where `viewpointOrderText` is defined, but at least there is no harm in adding this.

